PR #5, opened 2016-05-17
brozzler-new-site: Fix warcprox_meta's default value and json import

ato:
Just making it so the script can actually run.

I think brozzler-new-site is still partly broken though, the worker crashes complaining that job_id is None. I guess one needs to be passed to Site(). Wasn't sure if it should take a job id as an argument or generate one randomly.

I have since realised that brozzler-new-job is probably what I should be looking at instead but had already patched this so figured I'd submit it anyway just in case. ;-)

### DIFF
--- a/bin/brozzler-new-site
+++ b/bin/brozzler-new-site
@@ -27,6 +27,7 @@ import re
 import rethinkstuff
 import warnings
 import requests
+import json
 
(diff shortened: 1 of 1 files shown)


nlevitt:
Thank you Alex! brozzler-new-site doesn't get much attention now because we don't use it. But I want it to work, so fixes are more than welcome. Else, eventually I will get to it.

PR #7, opened 2016-05-28
disable browser extensions

galgeek:
--disable-extensions keeps Chromium from installing Chrome extensions—lastpass was very troublesome until I added this.

There's a great list of Chromium command line switches here: http://peter.sh/experiments/chromium-command-line-switches/

I'm curious to know why brozzler uses --window-size=1100,900

I'm curious to know wh


nlevitt:
I picked `--window-size=1100,900` arbitrarily as a not very big but not crazy small standard browser window size. Of course the screenshots will be this size. I'm definitely open to discussing or changing that...

PR #14, opened 2016-10-01
obsolete ARI-5008 / ARI-5065 Noah's ExtraHTTPHeaders encoding fix

galgeek:
The main purpose of this change is to avoid `content-encoding: br` which we were getting from facebook, but is not supported by wayback.

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -460,7 +460,7 @@ __brzl_compileOutlinks(window).join(' ');
         self.send_to_chrome(method="Runtime.enable")
 
         headers = self.extra_headers or {}
-        headers['Accept-Encoding'] = 'gzip, deflate'
+        headers['Accept-Encoding'] = 'identity'
         self.send_to_chrome(
                 method="Network.setExtraHTTPHeaders",
                 params={"headers":headers})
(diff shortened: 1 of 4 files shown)


brozzler/browser.py
```diff
@@ -457,8 +457,11 @@ def _visit_page(self, websock):
         self.send_to_chrome(method="Debugger.enable")
         self.send_to_chrome(method="Runtime.enable")
 
-        if self.extra_headers:
-            self.send_to_chrome(method="Network.setExtraHTTPHeaders", params={"headers":self.extra_headers})
+        headers = self.extra_headers or {}
+        headers['Accept-Encoding'] = 'gzip, deflate'
```
nlevitt: I'm thinking let's change this to 

        headers['Accept-Encoding'] = 'identity'

Because I'm not sure "deflate" is supported by the various wayback implementations (even if it really means gzip[1](https://stackoverflow.com/questions/9170338/why-are-major-web-sites-using-gzip/9186091#9186091))
And of course warcs are usually gzipped anyway, so gzipping inside of them doesn't save much space, and makes it more of a pain to inspect the contents.

PR #24, opened 2016-10-19
Use warcprox if enable_warcprox_features is true

mouse-reeve:


nlevitt:
A more complete description of this feature is, if `enable_warcprox_features` is true (for a Site, or set from the command line) but no proxy is set, choose an available instance of warcprox from the service registry in rethinkdb

PR #25, opened 2016-10-20
Add a mouseovers behavior

ato:
PR as requested. Although this literally is just a copy paste and string replace of 'click' with 'mouseover'.

Let me know if you'd like me to rework this in some other fashion. I did at first consider adding mouseover to simpleclicks.js rather than duplicating it but I realised I was just going to get bogged down trying figure out the most optimal approach.

It seems like a generic behavior that accepts a list of events + selectors might be a useful thing to have. Something like this:

``` yaml
events:
  - mouseover: .menu-item a
  - click: .tab
  - focus: input
  - click: button.dostuff
  - keypress:
      selector: .scrollything
      key: ArrowDown
```

But I haven't worked through enough examples yet to be confident of what's useful and what's premature generalisation.

### DIFF
--- a/brozzler/behaviors.d/mouseovers.js.template
+++ b/brozzler/behaviors.d/mouseovers.js.template
@@ -1,0 +1,136 @@
+/*
+ * brozzler/behaviors.d/mouseovers.js.in - mouseovers behavior template,
+ * mouseovers on elements matching templatized css selector
+ *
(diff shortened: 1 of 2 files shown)


nlevitt:
Thanks Alex. Merged this for now. I wonder how many of our existing behaviors could be generalized with the scheme that you propose. Definitely seems to make sense.

PR #26, opened 2016-10-29
Update Flash plugin instructions

ato:
libpepflashplayer.so is no longer included in the Chrome release package.  Adobe are [resuming Linux releases](http://blogs.adobe.com/flashplayer/2016/08/beta-news-flash-player-npapi-for-linux.html) and the PPAPI plugin is available from their [download site](https://get.adobe.com/flashplayer/otherversions/).


ato:
(CI failure is unrelated.)

PR #36, opened 2017-04-21
safen up brozzler.thread_raise() to avoid interrupting rethinkdb tran…

nlevitt:
…sactions and such

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -427,6 +427,7 @@ class Browser:
             self.websock_thread.on_request = on_request
         if on_response:
             self.websock_thread.on_response = on_response
+        brozzler.thread_accept_exceptions()
         try:
             self.navigate_to_page(
                     page_url, extra_headers=extra_headers,
@@ -444,13 +445,16 @@ class Browser:
             final_page_url = self.url()
             return final_page_url, outlinks
         except brozzler.ReachedLimit:
+            brozzler.thread_block_exceptions()
             # websock_thread has stashed the ReachedLimit exception with
             # more information, raise that one
             raise self.websock_thread.reached_limit
         except websocket.WebSocketConnectionClosedException as e:
+            brozzler.thread_block_exceptions()
             self.logger.error('websocket closed, did chrome die?')
             raise BrowsingException(e)
         finally:
+            brozzler.thread_block_exceptions()
             self.is_browsing = False
             self.websock_thread.on_request = None
             self.websock_thread.on_response = None
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -427,34 +427,31 @@ class Browser:
             self.websock_thread.on_request = on_request
         if on_response:
             self.websock_thread.on_response = on_response
-        brozzler.thread_accept_exceptions()
         try:
-            self.navigate_to_page(
-                    page_url, extra_headers=extra_headers,
-                    user_agent=user_agent, timeout=300)
-            if password:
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -32,7 +32,7 @@ def find_package_data(package):
 
 setuptools.setup(
         name='brozzler',
-        version='1.1b11.dev237',
+        version='1.1b11.dev238',
```
bnewbold: Incrementing this seems like an anti-pattern. What if two PRs both increment it at the same time? At least splitting to a separate commit would be cleaner?
Maybe this is established practice through; I don't have an alternative to propose without more context.

brozzler/browser.py
```diff
@@ -427,6 +427,7 @@ def browse_page(
             self.websock_thread.on_request = on_request
         if on_response:
             self.websock_thread.on_response = on_response
+        brozzler.thread_accept_exceptions()
```
bnewbold: Would definitely split the implementation and use of this new feature separate commits in the future.

tests/test_units.py
```diff
@@ -185,3 +192,67 @@ def test_start_stop_backwards_compat():
...
+            thread_caught_exception = e
+        finally:
+            logging.info('finishing')
+            let_thread_finish.clear()
+            thread_preamble_done.clear()
+
+    # test that thread_raise does not raise exception in a thread that has not
+    # called thread_accept_exceptions
+    thread_caught_exception = None
+    th = threading.Thread(target=lambda: thread_target())
```
bnewbold: I was going to recommend that this be just `threading.Thread(target=thread_target)`, but it's better and more consistent this way.

nlevitt:
> There are >80 lines of threading specific stuff in __init__.py across a couple functions... time to split into a utility file?

I could... for now __init__.py is my utility file, it still hasn't crossed my mental threshold for own fileness.

> Using a with block

Refactored to use a `with` block!

> extending the Thread class would put all this code in a single coherent place.

I decided not to do this because it would be more brittle. For example if some project is using brozzler as a library, and they want to call some brozzler code that accepts exceptions, they have to make sure it runs in a SpecialThread. They wouldn't be able to run such code at all from the main thread (unless they monkey patch it, yuck). Or on the flip side, if someone wants to use the `thread_raise` functionality outside of brozzler, perhaps in the main thread, or in threads created deep in some library code, there is no barrier to doing so.

> Incrementing this seems like an anti-pattern. What if two PRs both increment it at the same time?

(about incrementing the version number) Yeah I normally do that only on commits to master. Oops, not exactly sure how to pull it out of this pull request now.

> Would definitely split the implementation and use of this new feature separate commits in the future.

Oh? Well, I did that for you for this refactoring.

bnewbold:
After discussing in person, I think I understand the use case of this code a lot better. Pattern of sending exceptions to expedite shutdowns during specific "slow" code paths still feels a little hack-y to me. We discussed using multi-processing and signals as a way to pre-empt larger library code (like youtube-dl). For browser interactions, it might also be possible to signal the browser process directly (bypassing the owner thread) to expedite shutdown. One of the sign to me that multiprocessing would be acceptable is that "the amount of work is large/long", so the over head of a `fork` is negligable. Not sure either of those are really cleaner than this approach though.

Overall, this PR feels like an improvement to me, and particularly makes the code more understandable to future readers. "LGTM"!

nlevitt:
Thanks for looking it over @bnewbold!

PR #49, opened 2017-07-06
WIP—supports rewritten www.news.com.au yaml; good candidate for 2+ behaviors?

galgeek:
Updated interval function to read _actions_ and loop through them.

Updated www.news.com.au yaml for the new interval function. `brozzle-page http://www.news.com.au/` looks good.

I'm not sure we have a great candidate for multiple actions. I've updated the huffposts yaml—`brozzle-page` gets to the first page of slides and quits. However, this site includes a page load and URLs with hashtags; maybe we can come up with a less complex test case?

A couple of features to add soon, with additional properties in yaml: indexing _done_ items by target innerHTML rather than element, and closing elements we've opened—these would enable us to replace the new pm-gc-ca custom behavior, and then the new pitchfork behavior with closing.

### DIFF
--- a/brozzler/js-templates/umbraBehavior.js.j2
+++ b/brozzler/js-templates/umbraBehavior.js.j2
@@ -24,8 +24,9 @@ var umbraBehavior = {
     intervalId : null,
     state : null,
 
-    actions : [{'selector': 'div.teaser, li.pager__item a'}],
-    //actions : {{actions|json}},
+    //actions : [{'selector': 'div.teaser, li.pager__item a'}],
+    //actions : [{'selector': '.slideshow-card__overlay'}, {'selector': '.slideshow__next'}],
+    actions : {{actions|json}},
     k : 0,
 
     intervalFunc: function() {


galgeek:
The latest commit also appears to run properly with http://www.pm.gc.ca/eng/news/releases for its updated description in behaviors.yaml, limited to 4 clicks on the page’s _load more_ button.

brozzler/js-templates/umbraBehavior.js.j2
```diff
@@ -0,0 +1,174 @@
...
+                        somethingLeftBelow = true;
+                    } else if (where < 0) {
+                        somethingLeftAbove = true;
+                    }
+                    if (didSomething) {
+                        break; // break from doTargets loop, but not from documents loop
+                    }
+                }
+            }
+
```
nlevitt: 👇 should this section be outside of the `for (var k = 0; k < actionsLength; k++)` loop?

galgeek:
superseded by #71

PR #52, opened 2017-07-31
Remove redundant method parameter.

vbanos:
``ignore_cert_errors`` is passed to ``Chrome`` via ``Browser`` via ``BrowserPool` here:
https://github.com/internetarchive/brozzler/blob/master/brozzler/worker.py#L120

it is not doing anything in ``Browser.browser_page``.


nlevitt:
Thanks.

PR #54, opened 2017-08-01
Pass missing on_request callback to BrozzlerWorker methods

vbanos:
``Browser.browser_page`` has the ``on_request`` parameter but it is on
used by ``BrozzlerWorker._browse_page`` where we invoke it.

I add this to ``BrozzlerWorker.brozzle_page`` and pass it also to
``BrozzlerWorker._browser_page``. Now, it is possible to use this
callback from other applications when calling
``BrozzlerWorker.brozzle_page``.

### DIFF
--- a/brozzler/worker.py
+++ b/brozzler/worker.py
@@ -317,7 +317,7 @@ class BrozzlerWorker:
         return full_jpeg, thumb_jpeg
 
     def brozzle_page(self, browser, site, page, on_screenshot=None,
-                     enable_youtube_dl=True):
+                     on_request=None, enable_youtube_dl=True):
         self.logger.info("brozzling {}".format(page))
         if enable_youtube_dl:
             try:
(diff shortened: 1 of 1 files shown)


nlevitt:
(As mentioned on slack) I'm thinking that instead of adding only `on_request` to `brozzle_page`, we should add `**kwargs` and pass that all the way through to `browser.browse_page`.

vbanos:
You mean to eliminate ``_browse_page``? 
Looking at its contents, I see that it runs some extra logic before invoking ``browser.browse_page``.
You would miss that...

If I need to run ``browser.browse_page`` in my 3rd party software, I could invoke that. NP.

nlevitt:
> If I need to run browser.browse_page in my 3rd party software, I could invoke that. NP.

Yeah that's what I meant. Then you don't have to mess with Sites and Pages that you're not really using for anything. And you wouldn't need the enable_youtube_dl option either, but no problem with keeping it. Not sure if anything becomes more difficult this way, starting the browser maybe...

vbanos:
Yes, but what if I want to use warcproxy or youtube-dl?
Its not my case but it could be useful for other users.
In fact, maybe if we switch SPN to warcprox I'll definitely need ``brozzle_page``.

Also, as I see it the ``brozzle_page`` method is the "public" api
whereas the others are "internal" and could change in the future.

Do you need me to do any other change in the code? Thanks

nlevitt:
Yeah ok you would want to use `BrozzlerWorker.brozzle_page` to get the automatic saving of screenshots and so forth with warcprox. Ok I'll merge this. (Fwiw I would consider `Browser.browse_page` to also be public api.)

PR #56, opened 2017-08-03
Log oulinks for all users of Browser.

BitBaron:
This makes is such that anyone using the Browser class (e.g. Umbra) will see the outlinks in their logs.

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -452,6 +452,7 @@ class Browser:
                     outlinks = []
                 else:
                     outlinks = self.extract_outlinks()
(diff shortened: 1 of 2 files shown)


nlevitt:
What's the reason for this change?

BitBaron:
To make it so that users of the Browser class can few outlinks discovered. In this case, Umbra will log the outlinks as well.

nlevitt:
I don't think we want all this logged in brozzler-worker. Gonna decline.

nlevitt:
I'd rather not do that either. I thought it would be useful to print the outlinks that way for brozzle-page, but I wouldn't normally use logging for this kind of output (many lines without the timestamp prefix etc)

PR #62, opened 2017-08-17
update Instagram selectors

galgeek:
## Motivation
ARI-5460

## Description
Update selectors and click to dismiss new sign-in pop-up


BitBaron:
Ran local tests and all looked good. Thanks, @galgeek!

PR #65, opened 2017-08-31
Make behavior_timeout configurable

vbanos:
``behavior_timeout`` is hardcoded to 900s. With this MR we make it
configurable with a default value of 900. We add a new variable to
``BrozzlerWorker`` and ``Browser``.


galgeek:
Thank you, @vbanos!

PR #67, opened 2017-09-13
skip_youtube_dl

galgeek:
## Motivation
sometimes we want to skip youtube_dl...

## Testing
Tested so far with command line utility `brozzle-page`

### DIFF
--- a/brozzler/cli.py
+++ b/brozzler/cli.py
@@ -193,10 +193,9 @@ def brozzle_page(argv=None):
     browser = brozzler.Browser(chrome_exe=args.chrome_exe)
     try:
         browser.start(proxy=args.proxy)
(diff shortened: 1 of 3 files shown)


brozzler/cli.py
```diff
@@ -189,8 +193,10 @@ def on_screenshot(screenshot_png):
     browser = brozzler.Browser(chrome_exe=args.chrome_exe)
     try:
         browser.start(proxy=args.proxy)
+        enable_youtube_dl = False if args.skip_youtube_dl else True
```
nlevitt: This variable and the ternary conditional seem unnecessary... instead why not pass `enable_youtube_dl=not args.skip_youtube_dl` to brozzle_page?

galgeek: agreed, and updated—thanks, Noah!

PR #68, opened 2017-10-12
fix resume_job

danielbicho:
fix resume_job function, the job was not able to resume because the job stop_requested value was not reset.

## Motivation
After stopping a brozzler job with the brozzler-stop-crawl, calling the function resume_job was not able to resume the crawl. Altought resume job cannot work because of other job stop conditions, with a normal workflow like: brozzler-stop-crawl ---> resume_job will work now.

### DIFF
--- a/tests/test_frontier.py
+++ b/tests/test_frontier.py
@@ -287,7 +287,7 @@ def test_resume_job():
     # simulate job resume after a stop request
     frontier.resume_job(job)
     site1 = list(frontier.job_sites(job.id))[0]
-    site2 = list(frontier.job_sites(job.id))[0]
+    site2 = list(frontier.job_sites(job.id))[1]
 
     assert job.status == 'ACTIVE'
     assert job.stop_requested is None
@@ -308,7 +308,7 @@ def test_resume_job():
     site1.save()
 
     # should not raise a CrawlStopped
-    frontier.honor_stop_request(site)
+    frontier.honor_stop_request(site2)
 
     frontier.finished(site1, 'FINISHED_STOP_REQUESTED')
     job.refresh()
--- a/tests/test_frontier.py
+++ b/tests/test_frontier.py
@@ -203,19 +203,18 @@ def test_resume_job():
     assert len(job.starts_and_stops) == 2
     assert job.starts_and_stops[1]['start']
     assert job.starts_and_stops[1]['stop']
-    assert job.starts_and_stops[1]['stop'] > job.starts_and_stops[0]['start']
+    assert job.starts_and_stops[1]['stop'] > job.starts_and_stops[1]['start']
     assert site.status == 'FINISHED'
     assert len(site.starts_and_stops) == 2
     assert site.starts_and_stops[1]['start']
     assert site.starts_and_stops[1]['stop']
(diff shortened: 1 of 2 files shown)


nlevitt:
Do you mind adding a few lines of code to test the change in test_resume_job() in test_frontier.py?

danielbicho:
The travis is failing setting up the infrastructure?

tests/test_frontier.py
```diff
@@ -215,6 +215,7 @@ def test_resume_job():
     site = list(frontier.job_sites(job.id))[0]
 
     assert job.status == 'ACTIVE'
+    assert job.stop_requested is None
```
nlevitt: I don't think job.stop_requested is ever set, so this doesn't really test anything? Correct me if I'm wrong

nlevitt:
> The travis is failing setting up the infrastructure?

Yeah looks that way :(

anjackson:
This may be because the install process is using `/tmp/`. I had similar issues with Heritrix builds dying because `/tmp/` was being wiped mid-build.

nlevitt:
cryptography 2.1.1 seems to be the issue. It was released Oct 12. I can reproduce attempting to install it locally on ubuntu.

tests/test_frontier.py
danielbicho: You are right, it was just a assert to make sure that however the code evolve in the future it will just break the test if somehow the stop_request is not None after a resume. xD

I updated the test using the same pattern that it was being used before, but now it simulates a stop request.

tests/test_frontier.py
```diff
@@ -230,12 +230,123 @@ def test_resume_job():
...
+    assert site1.starts_and_stops[0]['stop']
+    assert site1.starts_and_stops[0]['stop'] > site.starts_and_stops[0]['start']
+    assert site2.starts_and_stops[0]['start']
+    assert site2.starts_and_stops[0]['stop']
+    assert site2.starts_and_stops[0]['stop'] > site.starts_and_stops[0]['start']
+
+    # simulate job resume after a stop request
+    frontier.resume_job(job)
+    site1 = list(frontier.job_sites(job.id))[0]
+    site2 = list(frontier.job_sites(job.id))[0]
```
nlevitt: Looks like you made the same kind of mistake here that I had made and you fixed above! :) I.e. I think it should be
```
    site2 = list(frontier.job_sites(job.id))[1]
```

tests/test_frontier.py
```diff
@@ -230,12 +230,123 @@ def test_resume_job():
...
+    assert len(site2.starts_and_stops) == 2
+    assert site2.starts_and_stops[1]['start']
+    assert site2.starts_and_stops[1]['stop'] is None
+
+    # simulate a site stop request
+    site1.stop_requested = datetime.datetime.utcnow().replace(tzinfo=doublethink.UTC)
+    site1.save()
+
+    # should not raise a CrawlStopped
+    frontier.honor_stop_request(site)
```
nlevitt: Maybe you meant
```frontier.honor_stop_request(site2)```
here?

tests/test_frontier.py
danielbicho: xD

nlevitt:
Thanks!

PR #70, opened 2017-11-07
skip remembering dash manifests

galgeek:
## Motivation
a recent brozzler crawl captured 13000+ files of mimetype video/vnd.mpeg.dash.mpd and recorded them as videos. these files are text manifests of segmented video files; they don't themselves include video content.

## Description
this update checks for mimetype video/vnd.mpeg.dash.mpd and skips recording such files as videos

apart from more formal testing, we can test as we troubleshoot wayback replay of youtube's new page layout


brozzler/worker.py
```diff
@@ -247,14 +247,16 @@ def _remember_videos(self, page, ydl_spy):
         if not 'videos' in page:
             page.videos = []
         for txn in ydl_spy.transactions:
-            if (txn['response_headers'].get_content_type().startswith('video/')
+            content_type = txn['response_headers'].get_content_type()
+            if (content_type.startswith('video/')
+                    and content_type != 'video/vnd.mpeg.dash.mpd'
```
nlevitt: Could you please add a comment somewhere nearby explaining why this mimetype is excluded?

galgeek:
Thanks, @nlevitt!
I've added comments.

I'm wondering if we need something a bit more complex than `site6` in current tests and noting this URL might prove helpful in constructing a more realistic test page: https://msdn.microsoft.com/en-us/library/dn551368(v=vs.85).aspx

nlevitt:
I found the instructions here helpful
https://developer.mozilla.org/en-US/docs/Web/HTML/DASH_Adaptive_Streaming_for_HTML_5_Video
Although really all of that probably wasn't necessary. The real important part is 
```self.extensions_map['.mpd'] = 'video/vnd.mpeg.dash.mpd'```

PR #73, opened 2018-01-04
Configurable JS templates location

vbanos:
Brozzler has hard-coded the JS templates logic in  ``brozzler/behaviors.yaml``
and ``brozzler/js-templates/`` locations. With this change, you can use
the optional ``behaviors_dir`` ``browser.browse_page`` parameter to set a
custom location and use any potential JS behaviors.

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -77,8 +77,8 @@ def behaviors(behaviors_dir=None):
     import os, yaml, string
     global _behaviors
     if _behaviors is None:
-        cwd = behaviors_dir or os.path.dirname(__file__)
-        behaviors_yaml = os.path.join(cwd, 'behaviors.yaml')
+        d = behaviors_dir or os.path.dirname(__file__)
+        behaviors_yaml = os.path.join(d, 'behaviors.yaml')
         with open(behaviors_yaml) as fin:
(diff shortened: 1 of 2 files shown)


brozzler/__init__.py
```diff
@@ -68,29 +68,34 @@ def _logger_trace(self, msg, *args, **kwargs):
...
+
+    :param behaviors_dir: Directory containing `behaviors.yaml` and
+    `js-templates/`. Defaults to brozzler dir.
+    """
     import os, yaml, string
     global _behaviors
     if _behaviors is None:
-        behaviors_yaml = os.path.join(
-                os.path.dirname(__file__), 'behaviors.yaml')
+        cwd = behaviors_dir or os.path.dirname(__file__)
```
nlevitt: `cwd` is a misleading name for this variable... could use `d`

vbanos: OK, fixed.

nlevitt:
i think maybe we should have a better system for this but for now it's fine

(like maybe an environment variable to specify a custom behaviors.yaml that is used in addition to the default, and takes precedence... and make `behavior_js_template` be a path instead of magically looking in one place for the files)
(but the format of behaviors.yaml doesn't lend itself to this because the order of items matters)

PR #90, opened 2018-02-13
Add chromium CLI flags to improve capture performance

vbanos:
``--disable-background-timer-throttling`` and ``--disable-renderer-backgrounding``:
karma JS test runner uses these to improve chrome performance
https://github.com/karma-runner/karma-chrome-launcher/issues/123

``--disable-hang-monitor``: Suppresses hang monitor dialogs in renderer
processes. This may allow slow unload handlers on a page to prevent the
tab from closing, but the Task Manager can be used to terminate the
offending process in this case.

``--mute-audio``: obvious.

The following are part of google safe browsing features:
``--disable-client-side-phishing-detection``
``--safebrowsing-disable-auto-update``
``--safebrowsing-disable-download-protection``

Good reference with all chromium cli opts: https://peter.sh/experiments/chromium-command-line-switches/

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -156,11 +156,16 @@ class Chrome:
                 '--use-mock-keychain', # mac thing
                 '--user-data-dir=%s' % self._chrome_user_data_dir,
                 '--disable-background-networking',
+                '--disable-renderer-backgrounding', '--disable-hang-monitor',
+                '--disable-background-timer-throttling', '--mute-audio',
                 '--disable-web-sockets', '--disable-cache',
                 '--window-size=1100,900', '--no-default-browser-check',
                 '--disable-first-run-ui', '--no-first-run',
                 '--homepage=about:blank', '--disable-direct-npapi-requests',
                 '--disable-web-security', '--disable-notifications',
+                '--disable-client-side-phishing-detection',
+                '--safebrowsing-disable-auto-update',
+                '--safebrowsing-disable-download-protection',
                 '--disable-extensions', '--disable-save-password-bubble']
         if self.ignore_cert_errors:
             chrome_args.append('--ignore-certificate-errors')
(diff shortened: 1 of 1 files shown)


nlevitt:
`--disable-background-timer-throttling` and `--disable-renderer-backgrounding`
It's not at all clear to me what these options do. Also background timer throttling sounds like it could be something good for performance? Hard to say though without knowing more about it. That one is not mentioned on the link you pasted.

`--mute-audio` - sure, probably good for anyone running brozzler on their laptops.

`--disable-hang-monitor` - probably a good idea, not sure what would happen in brozzler if that dialog popped up, probably best to avoid it.

`--disable-client-side-phishing-detection`
`--safebrowsing-disable-auto-update`
`--safebrowsing-disable-download-protection`
Hmm interesting. We're liable to archive some malware if we do this. Which could be something we want to archive. But on the other hand, is probably something we want to protect wayback users from viewing. What is global wayback's approach/policy on malware?

vbanos:
I'm sorry, info on all options I mentioned including the first two: `--disable-renderer-backgrounding`
`--disable-background-timer-throttling` is also in the Slack msg I wrote to you in the #brozzler channel.

`--disable-renderer-backgrounding`
_Prevents Chromium from lowering the priority of invisible pages' renderer processes._
Its used in all the projects I saw running headless chrome, e.g. check out this here:
https://github.com/karma-runner/karma-chrome-launcher/issues/123

`--disable-background-timer-throttling`
_Disable task throttling of timer tasks from background pages._
https://peter.sh/experiments/chromium-command-line-switches/#disable-background-timer-throttling
This is also used extensively. E.g. https://github.com/N0taN3rd/Squidwarc/blob/master/lib/crawler/launcher.js#L27

Regarding the safebrowsing options, since Global Wayback policy is to archive everything,
its best to remove them. I already did that.

nlevitt:
Thanks @vbanos. I saw the very brief descriptions that you quote in the docs but didn't understand what they meant. Thinking more about it, I can guess vaguely what they might mean and why they would improve performance. Viz, maybe chromium has optimizations to improve overall performance by de-prioritizing rendering of pages in windows that are not currently on screen. We want to disable these options because we want the browsers to run at full speed all the time. I'll merge your PR on the assumption that this interpretation is correct.

PR #95, opened 2018-02-18
Make Browser._wait_for sleep time a varible

vbanos:
Useful to be able to tweak this value in other apps using `Browser`.


nlevitt:
Sure why not.

PR #98, opened 2018-03-04
Disable chromium sandbox to improve performance

vbanos:
Use `--disable-sandbox` to disable chromium sandboxed processes
(restrictive environment for improved security). We don't need this in
the Brozzler use case and the performance is improved without it.

Info on chromium sandbox:
https://chromium.googlesource.com/chromium/src/+/master/docs/design/sandbox.md

The idea came up to me after seeing:
https://github.com/kennethreitz/requests-html/blob/master/requests_html.py#L467

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -155,6 +155,7 @@ class Chrome:
                 '--remote-debugging-port=%s' % self.port,
                 '--use-mock-keychain', # mac thing
                 '--user-data-dir=%s' % self._chrome_user_data_dir,
+                '--no-sandbox',
                 '--disable-background-networking',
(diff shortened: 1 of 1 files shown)


nlevitt:
> We don't need this in the Brozzler use case

Hmm. Are you sure? Brozzler is a potential attack vector, especially since IA's save-page-now for example will blithely open any url anybody in the world types in. Maybe we should bring this up in #security...

anjackson:
FWIW, I'd prefer that this was at least made optional rather than fully disabling it, for exactly this reason.

vbanos:
Its better to close this.
Note that I wrote in #security but didn't give any response.

PR #109, opened 2018-06-15
update instagram behavior

galgeek:
## Motivation
ARI-5747

## Description
update selector for detail image


BitBaron:
Good catch but that class name still looks a little gnarly and likely to change soon. Can we try something a little less likely to be random?

My quick tests with something like this in the browse console seem to do result in the same lists.
`document.querySelectorAll("a>div[role='button']")`

galgeek:
Thanks, @BitBaron!  That selector looks like it captures the same, and it may well not need an update again so soon.

I've updated the PR.

PR #120, opened 2018-09-21
handle http auth better

galgeek:
## Motivation
AITFIVE-1529

## Description
brozzler currently times out on pages requiring http auth—instead, we'd rather stop the browser and call the page finished

test added

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -58,6 +58,28 @@ class ReachedLimit(Exception):
     def __str__(self):
         return self.__repr__()
 
+class PageInterstitialShown(Exception):
+    def __init__(self, http_error=None, warcprox_meta=None, http_payload=None):
+        import json
+        if http_error:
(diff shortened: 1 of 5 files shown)


nlevitt:
Thanks Barbara, this is looking good. Just one thing:

```
        with pytest.raises(brozzler.PageInterstitialShown):
            browser.browse_page('https://monitor.archive.org/cgi-bin/nagios3/status.cgi?hostgroup=38.wbgrp')
```

We shouldn't rely on a random-ish page out on the internet for the automated tests. That's a job for the embedded http server: https://github.com/internetarchive/brozzler/blob/7497b7e5ac8b1/tests/test_brozzling.py#L49
Maybe add a clause to `do_GET()` for the path /401 to respond with a 401, something like that.

galgeek:
closing due to travis foo in favor of pr 138

PR #137, opened 2018-11-15
AITFIVE-1529 http auth and test

galgeek:
## Motivation
AITFIVE-1529

## Description
brozzler currently times out on pages requiring http auth—instead, we'd rather stop the browser and call the page finished

test added


galgeek:
closed in favor of earlier PR #120

PR #141, opened 2018-11-27
fetch service worker script with proper headers

nlevitt:


nlevitt:
A test case would be nice to have, but who has time for that? :)

PR #148, opened 2019-02-06
Add disk cache options to Chrome

vbanos:
Add `Chrome` options `disk_cache_dir` and `disk_cache_size` which add chromium
options `--disk-cache-dir=<DIR>` and `--disk-cache-size=N` (bytes).

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -63,7 +63,7 @@ class Chrome:
     logger = logging.getLogger(__module__ + '.' + __qualname__)
 
     def __init__(self, chrome_exe, port=9222, ignore_cert_errors=False,
-                 disk_cache=None, disk_cache_size=None):
+                 disk_cache_dir=None, disk_cache_size=None):
         '''
         Initializes instance of this class.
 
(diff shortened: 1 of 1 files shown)


nlevitt:
We discovered that the `--disable-cache` chromium option doesn't exist anymore and probably never did what we thought it did.
http://web.archive.org/web/20140205102645/http://peter.sh/experiments/chromium-command-line-switches

So @vbanos this is what I think we should do:

1. get rid of `--disable-cache` in all cases
2. rename `--disk-cache` to `--disk-cache-dir`
3. make `--disk-cache-dir` only accept a path (str), not a boolean
4. decouple `--disk-cache-size` from `--disk-cache-dir`, so it is possible to set either or both

I'd like to go on the record saying that sharing disk cache between different instances of chromium running at the same time seems risky to me. You might not have seen any problems in testing, but that is not conclusive. You could easily imagine race conditions where one browser is writing to a disk cache entry while another is reading from it, and the latter gets an incomplete or otherwise corrupt file. I don't know anything really about chromium's disk cache implementation, so I'm just speculating here. But I would assume it does locking within a browser process, but not necessarily across different browser processes.

PR #152, opened 2019-05-11
Add headless chrome option

vbanos:
Using `--headless` chromium-browser option, we run all Brozzler without
requiring an X display.
https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md

Benchmarks are ~20% faster using `--headless`.

This option was introduced in chromium-browser 59 but was unstable. I've
tested it with v73.0.3683.86 and its running well.

With this PR, we add it as an option that is disabled by default.

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -62,7 +62,8 @@ def check_version(chrome_exe):
 class Chrome:
     logger = logging.getLogger(__module__ + '.' + __qualname__)
 
-    def __init__(self, chrome_exe, port=9222, ignore_cert_errors=False):
+    def __init__(self, chrome_exe, port=9222, headless=False,
+                 ignore_cert_errors=False):
         '''
         Initializes instance of this class.
 
@@ -71,11 +72,13 @@ class Chrome:
         Args:
             chrome_exe: filesystem path to chrome/chromium executable
             port: chrome debugging protocol port (default 9222)
+            headless: run in headless mode, without X windows (default: False)
             ignore_cert_errors: configure chrome to accept all certs (default
                 False)
         '''
         self.port = port
         self.chrome_exe = chrome_exe
+        self.headless = headless
         self.ignore_cert_errors = ignore_cert_errors
         self._shutdown = threading.Event()
         self.chrome_process = None
@@ -134,13 +137,12 @@ class Chrome:
                     cookie_location, exc_info=True)
         return cookie_db
 
-    def start(self, headless=False, proxy=None, cookie_db=None,
+    def start(self, proxy=None, cookie_db=None,
               disk_cache_dir=None, disk_cache_size=None):
         '''
         Starts chrome/chromium process.
 
         Args:
-            headless: run in headless mode, without X windows (default: False)
             proxy: http proxy 'host:port' (default None)
             cookie_db: raw bytes of chrome/chromium sqlite3 cookies database,
                 which, if supplied, will be written to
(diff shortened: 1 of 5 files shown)


vbanos:
From my research, I found that some sites try to block chromium headless because they want to block "bots". There is no standard way to detect headless. One way is the user agent which becomes like:
```
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/60.0.3112.50 Safari/537.36
```
I suggest to use a custom one without the `headless` term in production.

nlevitt:
Great, thanks! Next step: make `py.test --tb=native -v tests/test_brozzling.py` pass.

(Fixing tests on travis-ci has been on my todo list for a long time. Sorry about that. But `test_brozzling.py` passes when I run it locally on master, so we should try not to regress.)

vbanos:
The situation is a bit complex.

1. Tests pass if we use `--headless -- disable-gpu` instead of `--headless` except the ones involving `youtube-dl`. (I suspect something is going on with my `youtube-dl` v2019.4.30 for that).

2. In the chrome page about headless it says: 
_Right now, you'll also want to include the --disable-gpu flag if you're running on Windows. See crbug.com/737678._
https://developers.google.com/web/updates/2017/04/headless-chrome
That is why I didn't use `--disable-gpu` in the first place. It shouldn't be necessary in linux.

3. The root cause of the following test failing:
```
py.test --tb=native -vvv tests/test_brozzling.py::test_page_interstitial_exception -s
```
is that when running chromium with `--headless` only, the event `Page.interstitialShown` is NOT returned!
When using `--headless --disable-gpu`, it is returned as expected.
I think that I discovered a chromium bug.

4. We could just use `--headless --disable-gpu` but the problem is that chromium becomes SLOWER when using it so there is no point. Other people had the same problem, e.g. https://github.com/Codeception/CodeceptJS/issues/561

vbanos:
In `brozzler.browser` you have the following comment when handling `Page.interstitialShown`:
```
            elif message['method'] == 'Page.interstitialShown':
                # AITFIVE-1529: handle http auth
                # we should kill the browser when we receive Page.interstitialShown and
                # consider the page finished, until this is fixed:
                # https://bugs.chromium.org/p/chromium/issues/detail?id=764505
                self.logger.info('Page.interstialShown (likely unsupported http auth request)')
                brozzler.thread_raise(self.calling_thread, brozzler.PageInterstitialShown)
```
The issue is reported to be fixed. Does this change our handling of the event?

nlevitt:
Thanks for investigating this. What do you think we should do about the chromium `Page.interstitialShown` bug? Want to file a bug with them?

> https://bugs.chromium.org/p/chromium/issues/detail?id=764505
> The issue is reported to be fixed. Does this change our handling of the event?

@galgeek is working on that :)

vbanos:
I will file a bug for `Page.interstitialShown`.

vbanos:
I just tried using the latest beta `Google Chrome 75.0.3770.38 beta` in case the problem is fixed but its not.
(They suggest trying that before filing a bug).

vbanos:
I reported the bug here: https://github.com/ChromeDevTools/devtools-protocol/issues/168
It seemed more appropriate. There wasn't any reference to devtools at chromium bug report form: https://bugs.chromium.org/p/chromium/issues/entry

vbanos:
Just tested with google-chrome 76.0.3809.87-1 in ubuntu 16.04 in case this problem is resolved but still see the same issue :(

TheTechRobo:
Rebase?

TheTechRobo:
The bug report was closed due to being in the wrong place...

PR #163, opened 2019-09-16
Add support for Facebook ads library and fix closing

CorentinB:
This pull request add support for the Facebook ads Library available at: https://www.facebook.com/ads/library/

The behavior added is to open and close the detailed metrics of each ads.

In addition to that, I've also added a fix to the closing mechanism, a missing `this.` to a call of the method `doTarget` prevented Brozzler from using the selectors specified in `closeSelector`, in my testings, no closing selectors were used, ever.

I have also added a new argument to the `doTarget` method specifying if the target is an open or a close element, this new argument is used to prevent closing selectors from being added to the `alreadyDone` array, we do not want/need to keep track of the closing selectors used.

Thanks!

### DIFF
--- a/brozzler/js-templates/umbraBehavior.js.j2
+++ b/brozzler/js-templates/umbraBehavior.js.j2
@@ -154,9 +154,9 @@ class UmbraBehavior {
         // an element we want to open. No need to track
         // the close selectors
         if (type == "open") {
+            this.idleSince = null;
             this.alreadyDone.push(target);
         }
-        this.idleSince = null;
     }
 
     start() {
--- a/brozzler/js-templates/umbraBehavior.js.j2
+++ b/brozzler/js-templates/umbraBehavior.js.j2
@@ -43,6 +43,7 @@ class UmbraBehavior {
 
         var documents = [];
         documents[0] = document;
+
         var iframes = document.querySelectorAll("iframe");
         var iframesLength = iframes.length;
         for (var i = 0; i < iframesLength; i++) {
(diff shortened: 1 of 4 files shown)


brozzler/js-templates/umbraBehavior.js.j2
```diff
@@ -54,25 +55,27 @@ class UmbraBehavior {
...
             }
         }
+
         var documentsLength = documents.length;
         for (var j = 0; j < documentsLength; j++) {
             if (closeSelector) {
                 var closeTargets = documents[j].querySelectorAll(closeSelector);
-                if ((closeTargets.length > 0) &&
-                    (this.alreadyDone.indexOf(closeTargets[0]) === -1) &&
-                    (this.isVisible(closeTargets[0]))) {
```
nlevitt: Is there a reason to remove the `isVisible()` check here? Seems like we would want to keep that.

brozzler/js-templates/umbraBehavior.js.j2
```diff
@@ -54,25 +55,27 @@ class UmbraBehavior {
...
-                    doTarget(closeTargets[0], 'click');
+                for (var i = 0; i < closeTargets.length; i++) {
+                    this.doTarget(closeTargets[i], "click", "close");
```
nlevitt: This `type` parameter with values "open"/"close" that change the behavior of `doTarget()` is not the most elegant design imho. Instead, I would just replace this `doTarget()` call with `closeTargets[i].click()`. Unless we know of close buttons that require a mouseover first? But I doubt it. @galgeek?

galgeek: I like the change you suggest, @nlevitt. 
I don't remember checking whether a mouseover is necessary, with the other behavior I worked on using a closeSelector, and the site is no longer on the live web.

PR #164, opened 2019-09-19
Update macOS instructions for Chromium installation

CorentinB:
Add chromium installation to the macOS instructions.


nlevitt:
Most people have google chrome installed and brozzler should find it automatically. Don't want to make people think it's more trouble to install brozzler than it actually is.

dvanduzer:
I'd be happy with a different phrasing for this. Just something along the lines of "you must have Chrome installed, or follow these alternate, footnoted instructions for homebrew."

Would've saved me some time. :)

PR #168, opened 2019-09-27
Implement facebook.js with behaviors.yaml

CorentinB:
Using behaviors.yaml templating system to implement the current Facebook behavior.

Opening pictures and closing them, and closing the login popup when it show up.


nlevitt:
100% support this idea. This implementation doesn't do the same stuff that the old behavior does though. What's the thought process behind it?

CorentinB:
> 100% support this idea. This implementation doesn't do the same stuff that the old behavior does though. What's the thought process behind it?

I took a look at what the code was doing, what selector it was looking for etc, and it looks outdated to me because it was basically only opening pictures and closing them, so I reproduced it with behaviors.yaml. If you think that I dropped `a[href^="/browse/likes"]` or `div[class="phm pluginLikeboxStream"]` for example, yes, but I didn't find these selectors on FB pages, if they are still somewhere on Facebook, do you know where please?

nlevitt:
Ok cool, that all sounds fine. What about `UMBRA_THINGS_TO_SCROLL_SELECTOR` and all of that stuff? (Also any particular reason for the change from `*[rel="theater"]` to `a[rel="theater"]`?)

CorentinB:
It looks like the selector in `UMBRA_THINGS_TO_SCROLL_SELECTOR` is for FB widgets in 3rd party pages, so it will never be activated if the URL isn't facebook.com, I am not sure why it was there.. and I do not see the selector it's specifying on any FB page.

And I changed from `*` to `a` because during my testings, using `*` wasn't doing more than `a` but was opening multiple times the same pictures.

nlevitt:
Cool! Mainly I wanted to make sure the omission was done intentionally and thoughtfully. I'd like to get your eyes on this @galgeek when you have a minute?

galgeek:
@CorentinB, thanks! It'll be great if we can replace the custom facebook behavior with a config for `umbraBehavior` in `behaviors.yaml`. I wonder if you've checked for `facebook.js` selectors on both logged-in and not-logged-in user pages, as well as logged-in and not-logged-in group pages.  (We've noticed recently significant differences in public group pages vs. private group pages, too.)

I'd also suggest checking `facebook.js` on a few pages using `brozzle-page`.

CorentinB:
> @CorentinB, thanks! It'll be great if we can replace the custom facebook behavior with a config for `umbraBehavior` in `behaviors.yaml`. I wonder if you've checked for `facebook.js` selectors on both logged-in and not-logged-in user pages, as well as logged-in and not-logged-in group pages. (We've noticed recently significant differences in public group pages vs. private group pages, too.)
> 
> I'd also suggest checking `facebook.js` on a few pages using `brozzle-page`.

I did check what `facebook.js` is doing on various type of pages, I can't see anything else than opening and closing pictures happening. Maybe I miss something, if I do, can you guide me into reproducing what you can achieve to do more on your side please? Thanks!

galgeek:
@CorentinB, thanks again!

Before merging a behavior update, we'll want to do a couple of qa crawls of facebook seeds, in our qa environment.  I'll try to get those started this week.

galgeek:
@CorentinB, thanks again!  I've deployed your update in the qa environment and started a couple of test crawls.

CorentinB:
I did a lot of testings to compare this implementation with the actual one. The current implementation is slightly faster in the scrolling process, but with my recent testings with a shorter interval on this implementation, I finally reverted to my initial 500ms choice, while being a big slower than the current `facebook.js` behavior, it is much more stable in what I've seen.

Also, I noticed that while having it implemented, the current behavior doesn't open ALL available pictures on the page, only the first pictures, before the login popup appear. This login popup that appear after a bit of scrolling on a Facebook page makes it impossible for Brozzler to open any additional pictures. This is fixed in this PR by adding a selector for closing the login popup.

nlevitt:
Barbara pointed out that with this change it's much slower getting to the bottom of the page, and doesn't capture as many scrolldowns as a result. Part of that is the longer interval (are you sure that change is necessary?), but the bigger problem is that it keeps scrolling back up to click stuff that it thinks it has missed above. That's old code, not something you introduced. But, it is an issue that had been addressed in facebook.js, which doesn't scroll back up until after it has reached the bottom of the page. Could you make umbraBehavior.js work that way too?

CorentinB:
@nlevitt I am not sure "much" slower is right but yes, it's slower, one of the reason is that it opens every pictures, actually, facebook.js doesn't. Past the first few pictures, facebook.js can't open any other picture while scrolling, because the login popup block this action, this is also fixed in this implementation, by closing this login popup.

Super cool that @galgeek made this PR, I suppose that with it, it shouldn't be that slower than the initial implementation, while being able to capture way more data. I made it faster then reverted because for such an inconsistent website that is Facebook, I noticed multiple times bad behaviors, things not loading fast enough, with a faster interval. Thanks you too for your work on this btw!

nlevitt:
> Past the first few pictures, facebook.js can't open any other picture while scrolling, because the login popup block this action

I'm not sure that's true, it looks to me like the picture modals are popping up even though the login modal is open. But closing it is a good idea anyway, I'm on board with that.

nlevitt:
> I finally reverted to my initial 500ms choice, while being a bit slower than the current facebook.js behavior, it is much more stable in what I've seen.

What did you observe specifically that led you to the conclusion that 500ms is more stable?

CorentinB:
@nlevitt very strange, I tested with `brozzle-page https://facebook.com/internetnetarchive` on the current master branch, can you share what you did in your side so I can try too?

With less than 500ms, I observed moments where the page was scrolling, but the loading of new posts took more time than expected for some reasons, and Brozzler exited thinking it was done. But if this 500ms choice is an issue, we can do less, what I observed did not happen ALWAYS, it was just more of a secure choice, most of the time it works well with less than 500ms.

nlevitt:
> @nlevitt very strange, I tested with brozzle-page https://facebook.com/internetnetarchive on the current master branch, can you share what you did in your side so I can try too?

Here's a way you can confirm that the login box doesn't cause problems:
- open up https://facebook.com/internetnetarchive in chrome
- open developer tools
- scroll down until the login popup appears
- in the devtools console tab, run this javascript:
   ```
   document.querySelectorAll('a[rel="theater"], a[id="expanding_cta_close_button"]')[0].click()
   ```
This is pretty much exactly what the behavior is doing, and for me it works, the modal pops up and loads the photo (in front of the login box in fact).

But anyway, this is immaterial! It's easy to close that login modal so we may as well do it. Certainly doesn't impact capture negatively.

> With less than 500ms, I observed moments where the page was scrolling, but the loading of new posts took more time than expected for some reasons, and Brozzler exited thinking it was done

Hmm. The interval time shouldn't have much impact on the behavior deciding it has finished. It checks whether it has had nothing to do for 10 sec (`IDLE_TIMEOUT_SEC`). Checking every 500ms vs 300ms shouldn't make much difference.

Could you change the interval to 300ms? Then I think it will be ready to merge.

galgeek:
umbraBehavior's default interval is 300ms, so it should be ok simply to omit that config line from behaviors.yaml.

Note that the custom facebook behavior interval is 200ms.

CorentinB:
@nlevitt yes it works with the dev tools, anyway as you say it's immaterial! I removed the custom interval as suggested by @galgeek.

Thanks!

PR #171, opened 2019-10-08
Add option to capture full page screenshot

vbanos:
Add option `full_page` to `Browser.screenshot`. The default behavior
remains the same.
We get inspiration from puppeteer to capture a screenshot of the full
page:
https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L898

Add option `screenshot_full_page=False` to `Browser.browse_page` to use
the new feature when capturing a page.

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -616,9 +616,9 @@ class Browser:
                             deviceScaleFactor=deviceScaleFactor,
                             screenOrientation=screenOrientation)
                 )
-            capture_params = {'format': 'jpeg', 'clip': clip}
+            capture_params = {'format': 'jpeg', quality: 95, 'clip': clip}
         else:
-            capture_params = {'format': 'jpeg'}
+            capture_params = {'format': 'jpeg', quality: 95}
         self.websock_thread.expect_result(self._command_id.peek())
         msg_id = self.send_to_chrome(method='Page.captureScreenshot',
                                      params=capture_params)
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -487,8 +487,6 @@ class Browser:
                             'login navigated away from %s; returning!',
                             page_url)
                         self.navigate_to_page(page_url, timeout=page_timeout)
-                if on_screenshot:
-                    self._try_screenshot(on_screenshot, screenshot_full_page)
                 behavior_script = brozzler.behavior_script(
                         page_url, behavior_parameters,
                         behaviors_dir=behaviors_dir)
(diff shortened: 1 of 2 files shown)


vbanos:
Please also note the following:
The original `screenshot` code says `jpeg_bytes` which implies that it captures JPEG images.
`Page.captureScreenshot` defaults to PNG.
https://chromedevtools.github.io/devtools-protocol/1-3/Page#method-captureScreenshot

That's why in my code, I use `'format': 'jpeg'` in all cases.

vbanos:
Example screenshot:
brozzler-fullscreen

nlevitt:
> Please also note the following:
> The original `screenshot` code says `jpeg_bytes` which implies that it captures JPEG images.
> `Page.captureScreenshot` defaults to PNG.
> https://chromedevtools.github.io/devtools-protocol/1-3/Page#method-captureScreenshot
> 
> That's why in my code, I use `'format': 'jpeg'` in all cases.

👍 Yeah `jpeg_bytes` was incorrect, oops. But take a look at this code:
https://github.com/internetarchive/brozzler/blob/master/brozzler/worker.py#L229
It expects png bytes and actually converts to jpeg. This code needs to be adjusted to expect jpeg.

Separately,
```
quality integer
Compression quality from range [0..100] (jpeg only). 
```
https://chromedevtools.github.io/devtools-protocol/tot/Page#method-captureScreenshot

Do you know what chrome uses as the default value for `quality`? Not documented here. Brozzler is using 95 in `BrozzlerWorker .full_and_thumb_jpegs()`. It might make sense to send 95 in the Page.captureScreenshot call, but it would be good to know what the default is to inform that decision.

vbanos:
@nlevitt so I need to move `on_screenshot` callback and make it run after behavior runs. I'm on it.

Another cause for this problem could be a very small JS behavior timeout. In SPN2, I use just 5 sec because we need to be very fast to handle the load.

vbanos:
After moving the screenshot code to run right after the JS behavior and increasing the JS behavior timeout from 5 to 15 sec, I was able to take a nice screenshot that includes all images.

I used both PNG and JPEG format and I see that PNG size is 5.8 MB whereas JPEG size is 1.3 MB !! We should definitely use JPEG.
new

vbanos:
The default JPEG quality is 80. I found out using ImageMagic `identify` on the created JPEG screenshot.

nlevitt:
I tried this out a bit and I think it's all good.

> the whole thing with Page.getLayoutMetrics is kind of weird and interesting
> now i understand what that’s for and… it’s probably good?
> hrm even jpeg with quality=100 is artifact-y around this link text
> i guess we have to live with that
> why did you decide to scroll to the top before taking the screenshot?
> ah i think i see why. i tried 'clip': dict(x=0, y=0, width=1100, height=900, scale=1) for the non-full page screenshot and ended up with a blank (all white) screenshot

(Damn, some of these BBC headlines read like clickbait)

PR #173, opened 2019-10-16
Run JS behaviors only on HTML

vbanos:
Make method `Browser.page_is_html` which return boolean. The method uses
`document.contentType` to get current page's conent-type and decide if its HTML.

If page is not HTML, skip JS behavior.

Capture performance increases greatly for non html pages.

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -655,7 +655,24 @@ class Browser:
         message = self.websock_thread.pop_result(msg_id)
         return message['result']['result']['value']
 
+    def page_is_html(self, timeout=10):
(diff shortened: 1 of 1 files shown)


nlevitt:
Hrm, I'm not real excited about doing this extra call for every page, when in the normal brozzler case, we already know it's html, when we're in this section of code. What if instead in SPN you make a HEAD request first, and if it's not html, do a simple fetch (maybe using `BrozzlerWorker._fetch_url()`)?

vbanos:
OK, I can do that, thanks.

PR #181, opened 2019-11-15
Enable running in docker / k8s

vbanos:
When trying to run Brozzler in docker, we get the following error:
```
Failed to move to new namespace: PID namespaces supported, Network
namespace supported, but failed: errno = Operation not permitted
Trace/breakpoint trap
```
This happens because chromium uses sandboxing for increased security by
default and its not supported when running in a container.
Adding chromium option `--no-sandbox` fixes the problem.

This issue is common, I found various reports about it like this:
https://github.com/Zenika/alpine-chrome/issues/33

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -170,7 +170,7 @@ class Chrome:
                 '--disable-background-networking',
                 '--disable-renderer-backgrounding', '--disable-hang-monitor',
                 '--disable-background-timer-throttling', '--mute-audio',
-                '--disable-web-sockets',
+                '--disable-web-sockets', '--no-sandbox',
                 '--window-size=1100,900', '--no-default-browser-check',
                 '--disable-first-run-ui', '--no-first-run',
                 '--homepage=about:blank', '--disable-direct-npapi-requests',
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -179,7 +179,7 @@ class Chrome:
 
         extra_chrome_args = os.environ.get('BROZZLER_EXTRA_CHROME_ARGS')
         if extra_chrome_args:
-            chrome_args.append(extra_chrome_args)
+            chrome_args.extend(extra_chrome_args.split())
         if disk_cache_dir:
             chrome_args.append('--disk-cache-dir=%s' % disk_cache_dir)
         if disk_cache_size:
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -170,13 +170,16 @@ class Chrome:
                 '--disable-background-networking',
                 '--disable-renderer-backgrounding', '--disable-hang-monitor',
                 '--disable-background-timer-throttling', '--mute-audio',
-                '--disable-web-sockets', '--no-sandbox',
+                '--disable-web-sockets',
                 '--window-size=1100,900', '--no-default-browser-check',
                 '--disable-first-run-ui', '--no-first-run',
                 '--homepage=about:blank', '--disable-direct-npapi-requests',
                 '--disable-web-security', '--disable-notifications',
                 '--disable-extensions', '--disable-save-password-bubble']
 
+        extra_chrome_args = os.environ.get('BROZZLER_EXTRA_CHROME_ARGS')
+        if extra_chrome_args:
+            chrome_args.append(extra_chrome_args)
         if disk_cache_dir:
             chrome_args.append('--disk-cache-dir=%s' % disk_cache_dir)
         if disk_cache_size:


vbanos:
I have some other observations when trying to run chromium / Brozzler in docker.
chromium tries to use d-bus but its not available so it reports errors like the following. Its working fine though. Maybe I need to find out how to enable d-bus in docker.
```
1035:1272:1115/130216.784860:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[1035:1272:1115/130216.784967:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[1035:1272:1115/130216.785068:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[1035:1272:1115/130216.785221:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[1035:1272:1115/130216.785278:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
```

vbanos:
Even though we already use chromium option `--ignore-certificate-errors`, we still get errors like the following:
```
[1064:1083:1115/130216.794595:ERROR:cert_verify_proc_nss.cc(1011)] CERT_PKIXVerifyCert for accounts.google.com failed err=-8179
[1064:1083:1115/130216.818862:ERROR:cert_verify_proc_nss.cc(1011)] CERT_PKIXVerifyCert for archive.org failed err=-8179
[1064:1083:1115/130217.578483:ERROR:cert_verify_proc_nss.cc(1011)] CERT_PKIXVerifyCert for analytics.archive.org failed err=-8179
[1064:1083:1115/130217.691415:ERROR:cert_verify_proc_nss.cc(1011)] CERT_PKIXVerifyCert for clients1.google.com failed err=-8179
[1064:1083:1115/130226.727836:ERROR:cert_verify_proc_nss.cc(1011)] CERT_PKIXVerifyCert for ssl.gstatic.com failed err=-8179
```

vbanos:
chromium cannot run as `root` but `root` is the default docker user. You need to do something like the following before launching it:
```
RUN groupadd -r spn && useradd -r -g spn spn                                    
USER spn
```

vbanos:
The full command I used for testing is what Brozzler uses:
```
/usr/bin/chromium-browser --remote-debugging-port=33839 --use-mock-keychain --user-data-dir=/dev/shm/tmp62kztegu/chrome-user-data --disable-background-networking --disable-renderer-backgrounding --disable-hang-monitor --disable-background-timer-throttling --mute-audio --disable-web-sockets --window-size=1100,900 --no-default-browser-check --disable-first-run-ui --no-first-run --homepage=about:blank --disable-direct-npapi-requests --disable-web-security --disable-notifications --disable-extensions --disable-save-password-bubble --ignore-certificate-errors --proxy-server=http://wwwb-spn10:8000 --no-sandbox http://archive.org
```
If you remove `--no-sandbox` it crashes with the error I already mentioned in this PR.

nlevitt:
Is there a way we can support docker without --no-sandbox? 

Maybe you could investigate these leads?

https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#setting-up-chrome-linux-sandbox says
```
Add user so we don't need --no-sandbox.
```

https://github.com/Zenika/alpine-chrome#3-ways-to-use-chrome-headless-with-this-image
has this little section:
```
The best: With seccomp
```

vbanos:
When I'm trying to follow the instructions on the first link you posted and run:
`sudo sysctl -w kernel.unprivileged_userns_clone=1`
I get an error because of a read only filesystem.
Docker needs to run with some extra conf to allow this. https://docs.docker.com/engine/security/userns-remap/
I think that there will be issues when trying to run this in our k8s clusters.

vbanos:
When trying to use the instructions on the 2nd link, I get another error:
`The setuid sandbox provides API version 1, but you need 0`
`Please read` https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md.

vbanos:
@nlevitt I have two alternative suggestions:
1. When ENV var `DISABLE_CHROME_SANDBOX` is set, we enable `--no-sandbox`
2. Use ENV var `EXTRA_CHROME_ARGS` to pass arbitrary cli arguments to chrome.

I think that the 2nd would be better because maybe we'll need to use other options as well soon. E.g. some time ago I told you about `--disable-dev-shm-usage` to use /tmp for all temp files and how this is useful when running in docker. https://bugs.chromium.org/p/chromium/issues/detail?id=736452

nlevitt:
@vbanos I like #2, let's do that. Maybe call it `BROZZLER_EXTRA_CHROME_ARGS`

vbanos:
Its ready!

brozzler/chrome.py
```diff
@@ -177,6 +177,9 @@ def start(self, proxy=None, cookie_db=None, disk_cache_dir=None,
                 '--disable-web-security', '--disable-notifications',
                 '--disable-extensions', '--disable-save-password-bubble']
 
+        extra_chrome_args = os.environ.get('BROZZLER_EXTRA_CHROME_ARGS')
+        if extra_chrome_args:
+            chrome_args.append(extra_chrome_args)
```
nlevitt: Hmm, I think you're gonna need to split that on white space in case `BROZZLER_EXTRA_CHROME_ARGS='--arg1 --arg2'`

vbanos: I've checked this specific case and its no problem.
If you use `--arg1 --arg2` it will be added to `chrome_args` like all the rest of the parameters and the final chrome command will have `--arg1 --arg2`.

Initially, I considered doing:
```
if extra_chrome_args:
    chrome.args.extend(extra_chrome_args.split(' '))
``
but its exactly the same result and the one we'll have now and the current code is simpler.

Comment (nlevitt) [reply brozzler/chrome.py]: You need the split. Also you should just do `extra_crome_args.split()` to handle any kind of whitespace.

Compare:
```
>>> import subprocess
>>> subprocess.Popen(['ls', '-a', '-l'])
<subprocess.Popen object at 0x10dcc03d0>
>>> total 0
drwxr-xr-x   4 nlevitt  wheel  128 Nov 27 10:34 .
drwxrwxrwt  15 root     wheel  480 Nov 27 10:34 ..
-rw-r--r--   1 nlevitt  wheel    0 Nov 27 10:34 .a
-rw-r--r--   1 nlevitt  wheel    0 Nov 27 10:34 b
```
```
>>> subprocess.Popen(['ls', '-a -l'])
<subprocess.Popen object at 0x10dcc08d0>
>>> ls: illegal option --  
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

nlevitt:
Thanks!

PR #182, opened 2019-11-25
Fix Facebook ads variant selector

CorentinB:


nlevitt:
I took at look at this page and I see why we have to use these horrible css selectors.  But I'm gonna suggest this one which I think is a little more elegant and perhaps less fragile:
```div._7plx button:last-child```

Incidentally ```i[class="_271o img sp_KBE8sh--02o sx_5d0205"]``` can be written more idiomatically as ```i._271o.img.sp_KBE8sh--02o.sx_5d0205```.

CorentinB:
@nlevitt thanks! Did you try this selector on Brozzler directly? It doesn't work on my side. I can use it to find the button via the dev tools on Chrome of course, but it doesn't work on Brozzler, as the `childSelector`.

nlevitt:
Huh you're right. It seems to be the first click on "See Ad Details" that isn't working. I have zero explanation for this right now. Oh well

PR #190, opened 2020-03-23
Skip running behaviors when page is 4xx or 5xx

vbanos:
Currently, when we run `Browser.browse_page`, we run JS behaviors after
we navigate to a page regardless of its status.
Maybe the page wasn't found (4xx) or unreachable for any reason (5xx).
In that case, we could skip running behaviors to save time and
resources.

With this PR, we add a new var to store navigated page HTTP status in
`WebsockReceiverThread.page_status`. We use this in
`Browser.browser_page` to skip behaviors, outlink and hashtag extraction
when page status is 4xx/5xx.

Note that we don't skip screenshots as it could be useful to have a
picture of an error page in some cases.

Finally, we add option `simpler404` to enable this (it is disabled by default).

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -422,8 +422,8 @@ class Browser:
             on_service_worker_version_updated=None, on_screenshot=None,
             username=None, password=None, hashtags=None,
             screenshot_full_page=False, skip_extract_outlinks=False,
-            skip_visit_hashtags=False, skip_youtube_dl=False, page_timeout=300,
-            behavior_timeout=900):
+            skip_visit_hashtags=False, skip_youtube_dl=False, simpler404=False,
+            page_timeout=300, behavior_timeout=900):
         '''
         Browses page in browser.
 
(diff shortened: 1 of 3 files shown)


brozzler/browser.py
```diff
@@ -494,19 +498,29 @@ def browse_page(
...
                 if on_screenshot:
                     self._try_screenshot(on_screenshot, screenshot_full_page)
-                if skip_extract_outlinks:
+                if not run_behaviors or skip_extract_outlinks:
                     outlinks = []
                 else:
                     outlinks = self.extract_outlinks()
-                if not skip_visit_hashtags:
-                    self.visit_hashtags(self.url(), hashtags, outlinks)
-                final_page_url = self.url()
```
galgeek: do we want to omit line 509?

vbanos: I moved line 509 to 515 in the new code.

The reason is that in the previous code, we use `self.url()` 2 times a) `self.visit_hashtags(self.url(),...` and b) `final_page_url = self.url()`.
I wanted to use it only once because its slow (it contacts the browser to get the current URL via websocket).
The final page URL doesn't change when you are at that point of the method, so it seems redundant.

So, I execute `final_page_url = self.url()` on line 515 and then I use `final_page_url` instead of `self.url()`.

PR #193, opened 2020-04-13
Add unit tests for try_login behavior

vbanos:
Add unit tests for the code that detects and tries to use login forms
automatically (`Browser.try_login`).

Add `htdocs/favicon.ico` because it is loaded automatically when the
browser tries to use the test web server and it causes a "missing"
warning.

Create a new dir `tests/htdocs/site11` which is used for login related
test html files.

Our aim is to expand the auto login functionality but before doing that,
we need to have proper unit testing to be on the safe side.


tests/test_brozzling.py
```diff
@@ -72,6 +72,16 @@ def do_GET(self):
...
 
+        def do_POST(self):
+            if self.path == '/login-action':
+                self.send_response(200)
+                payload = b'login successfull\n'
+                self.send_header('Content-Type', 'text/plain;charset=utf-8')
+                self.send_header('Content-Length', len(payload))
+                self.end_headers()
+                self.wfile.write(payload)
+
```
galgeek, line 83: any reason not to add this?
```
else:
    super().do_POST
```

vbanos: I just missed it. Thank you! I've just added it.

PR #194, opened 2020-04-14
Expanding Brozzler's logging in capabilities

NGTmeaty:
Some sites don't allow you to login without clicking on a button to open a retracted modal.
This update to the login code allows Brozzler to click on all elements that we think are related to opening a login modal. Then, if there isn't a regular form, we will attempt to fill out abnormal form schemes. The test_try_login test has been expanded for the new type of login form we are supporting.

### DIFF
--- a/brozzler/js-templates/try-login.js.j2
+++ b/brozzler/js-templates/try-login.js.j2
@@ -1,8 +1,9 @@
 var __brzl_tryLoginState = 'trying';
 
 var __brzl_tryLogin = function() {
-    for(var x = 0; x < document.querySelectorAll("[class^='login-open']").length; x++){
-        button = document.querySelectorAll("[class^='login-open']")[x];
+    var login_open = document.querySelectorAll("[class^='login-open']");
+    for(var x = 0; x < login_open.length; x++){
+        button = login_open[x];
         button.click();
     }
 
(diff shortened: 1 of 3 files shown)


brozzler/js-templates/try-login.js.j2
```diff
@@ -1,6 +1,11 @@
 var __brzl_tryLoginState = 'trying';
 
 var __brzl_tryLogin = function() {
+    for(var x = 0; x < document.querySelectorAll("[class^='login-open']").length; x++){
+        button = document.querySelectorAll("[class^='login-open']")[x];
```
vbanos, line 5: You are running `querySelectorAll` too many times here, N times in the `for` loop and once more to get the `button`.
You could do `var login_open = document.querySelectorAll("[class^='login-open']")` and use that.

brozzler/js-templates/try-login.js.j2
```diff
@@ -48,6 +53,39 @@ var __brzl_tryLogin = function() {
             return;
         }
     }
+     
+
```
vbanos, line 57: Please avoid unnecessary empty lines, one empty line is fine to separate code parts.

brozzler/js-templates/try-login.js.j2
```diff
@@ -48,6 +53,39 @@ var __brzl_tryLogin = function() {
...
+    if(__brzl_tryLoginState === 'trying'){
+        var submit = undefined;
+        var usernameInput = undefined;
+        var passwordField = undefined;
+
+        for (var z = 0; z < document.querySelectorAll("[class^='login']").length; z++){
+            var input = document.querySelectorAll("[class^='login']")[z];
```
vbanos, line 64: My previous comment on too many `querySelectorAll` applies here as well.

PR #197, opened 2020-05-11
Login to paywalled site that uses a separate login page

vbanos:
Currently, Brozzler tries to login in a paywalled site if the login form is inside the target URL (`try_login` method).

The problem we are trying to resolve is that some times, the target URL is behind a paywall and you need to click the "Login" button to go to the login page, authenticate and then return to the target page.

We define a new `login-behaviors.yaml` where we can define rules to visit login pages for specific sites. We use as an example https://www.thetimes.co.uk.

We try to reuse existing code (`brozzler.behaviors`, `brozzler.behavior_script`).

Our code runs just before `try_login`. If the target site is defined in `login-behaviors.yaml`, we try to click on the "login" button (we find it using a selector) and then we run `try_login` as we already do.

The example we are using is not very good because `try_login` does not login successfully in https://account.thetimes.co.uk/ (we have valid credentials). This is the topic of another PR where we need to improve `try_login` to support it.

We are trying to make a correct login mechanism and then later add rules to `login-behaviors.yaml`.

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -111,12 +111,12 @@ def behaviors(behaviors_dir=None, conf='behaviors.yaml'):
             _behaviors = yaml.safe_load(fin)
     return _behaviors
 
-def behavior_script(url, template_parameters=None, behaviors_dir=None):
+def behavior_script(url, template_parameters=None, behaviors_dir=None, behaviors=None):
     '''
     Returns the javascript behavior string populated with template_parameters.
     '''
     import re, logging, json
-    for behavior in behaviors(behaviors_dir=behaviors_dir):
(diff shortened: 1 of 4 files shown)


vbanos:
This code doesn't work correctly due to JS behavior issues with the login form.

Most of the sites we are interested in archiving use https://auth0.com/ login system or something similar
to block logging in using bots / headless browsers.

We hit the login link/button, view the login form but when we try to insert credentials there, the target site
understands that we are a bot and blocks us from submitting the form and logging in.

This code worked with some trivial cases like accessing a page behind the `/wp-admin/` area of a wordpress site but we don't care about that :slightly_smiling_face:

PR #200, opened 2020-06-02
Attempt to fix the tests

NGTmeaty:
RethinkDB wasn't installing, due to [a change in download servers.](https://rethinkdb.com/blog/download-server-changes) The latest version of Numpy no longer supports Python 3.5, so to maintain backwards compatibility, when Python 3.5 is being used we install a older version.

I'm still looking into the main test failure I've currently seen, test_no_proxy not passing due to Travis timing it out. Thanks!


galgeek:
thanks, @NGTmeaty!

I'm checking your PR in brozzler's qa branch, where so far we've run into other issues, very possibly unrelated to your PR.

PR #207, opened 2020-08-06
Travis CI: Add Python 3.8

cclauss:
https://devguide.python.org/#status-of-python-branches
https://devguide.python.org/devcycle/#end-of-life-branches # Python 3.5 EOLs in one month.

Also, duplicate keys are not allowed so only one will be used...
Screenshot 2020-08-06 at 16 53 08

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,13 @@ os: linux
 dist: xenial
 language: python
 python:
+- 3.5
 - 3.6
 - 3.7
 - 3.8
+jobs:
+  allow_failures
+    - python: 3.8
 before_install:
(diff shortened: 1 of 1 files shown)


galgeek:
Thanks, @cclauss!

Do version numbers need quotes now?

cclauss:
Yes they do for allow_failures to work.  Much has been written about the problems of floating point numbers in yaml files.  Like the fact that 3.10 == 3.1 which is will soon be a problem now that Python 3.9fc1 has been released.

galgeek:
Thanks, @cclauss!

PR #208 supersedes this.

PR #209, opened 2020-10-04
Add option extract_outlinks_timeout

vbanos:
`Browser.extract_outlinks` has a default `timeout=60` parm that cannot be
changed in any way. (It is always invoked using `extract_outlinks()`.

We add param `extract_outlinks_timeout=60` to `BrozzlerWorker` and
`Browser.browse_page` to allow that.


vbanos:
This feature is necessary for some users, e.g. in SPN2 we need to spend less time on `extract_outlinks` and we want to use a smaller timeout value.

PR #210, opened 2020-10-06
Remove "proxy" references and fix scripts' paths

ldko:
While trying out Brozzler via the Vagrant setup provided in the repo, I hit a couple things that could possibly use updating. If it is of interest, I have included them here.

The PR removes references to "proxy" in the job-conf.rst and vagrant/README.rst, which, as far as I could tell, is no longer a valid field in configuring jobs, as it is also no longer a valid flag when calling `brozzler-new-site`.

This also updates the virtualenv path in ` vagrant/vagrant-brozzler-new-job.py` where it was no longer accurate and adds use of `os.path.realpath` to allow flexibility in the directory from which the vagrant-brozzler-* scripts are run.

### DIFF
--- a/job-conf.rst
+++ b/job-conf.rst
@@ -14,7 +14,6 @@ Example
 
     id: myjob
     time_limit: 60 # seconds
-    proxy: 127.0.0.1:8000 # point at warcprox for archiving
     ignore_robots: false
     max_claimed_sites: 2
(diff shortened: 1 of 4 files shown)


TheTechRobo (approved):
Should work - I tried removing proxy argument, and it still used warcprox. LGTM!

PR #211, opened 2020-10-13
configurable websocket url timeout, default 60

galgeek:
identical code tested in qa branch


vbanos:
Improvement suggestion:

`Chrome.start` should have an extra argument `websocket_timeout=60` which should be passed to `Chrome._websocket_url`
like this: `return self._websocket_url(timeout_sec=websocket_timeout)`.

Without this, there isn't a way to change the default websocket timeout.

galgeek:
Thanks, @vbanos — I've updated the PR.

PR #223, opened 2021-05-23
Skip invalid outlink

vbanos:
When one of the outlinks is `http://-1/` `urlcanon.whatwg` raises an
unhandled exception `ipaddress.AddressValueError` and the capture fails.

We can skip the problematic outlink and keep the rest without crashing.


vbanos:
You can replicate this if you try to capture http://musicmachinery.com/2009/04/27/moot-wins-time-inc-loses/

PR #246, opened 2022-06-17
Add stealth parameter to avoid antibot systems

vbanos:
The aim is to prevent Brozzler detection and blocking by antibot
systems. To do that, we need to run some JS before any other code runs
on page load and mock specific browser attributes which indicate that
Brozzler is a bot.

We add the option `stealth` in `Browser`, `brozzler.cli` and
`BrozzlerWorker`. It is disabled by default.

If enabled, we run `stealth.js` which is executed before anything else
on the page via `Page.addScriptToEvaluateOnNewDocument`.

For now, we mock only the graphics driver attributes.
If this is OK, we can add more antibot evasions in the same script.

There are many antibot tests, we are using this: https://bot.sannysoft.com/

Inspired mainly by:
https://www.npmjs.com/package/puppeteer-extra-plugin-stealth

### DIFF
--- a/brozzler/browser.py
+++ b/brozzler/browser.py
@@ -425,7 +425,7 @@ class Browser:
             screenshot_full_page=False, skip_extract_outlinks=False,
             skip_visit_hashtags=False, skip_youtube_dl=False, simpler404=False,
             page_timeout=300, behavior_timeout=900,
-            extract_outlinks_timeout=60, download_throughput=-1):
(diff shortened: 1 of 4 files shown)


vbanos:
Check out the "WebGL Vendor" and "WebGL Renderer" fields with and without stealth mode.
The mock values are from my desktop Ubuntu 22.04 using an NVIDIA Quadro graphics card.
Brozzler is running on a VM using XVNC.

screenshot-stealth

screenshot

PR #248, opened 2022-07-29
Add more stealth evasions

vbanos:
Set `navigator.platform = 'Win32'` instead of the default `Linux` as we
usualy run Brozzler on Linux.

Randomize the `navigator.deviceMemory` and
`navigator.hardwareConcurrency` to avoid browser fingerprinting.

Define `window.Notification` which is not defined because we run Chrome
with CLI parameter `--disable-notifications`.

### DIFF
--- a/brozzler/js-templates/stealth.js
+++ b/brozzler/js-templates/stealth.js
@@ -14,6 +14,34 @@ WebGLRenderingContext.prototype.getParameter = function(origFn) {
   };
 }(WebGLRenderingContext.prototype.getParameter);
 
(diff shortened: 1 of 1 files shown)


vbanos:
The `window.Notification` improvement is very important. Check out 2 screenshots of https://bot.sannysoft.com/ before and after using it.
Something is crashing in the browser detection library if it isn't there.

BEFORE:
initial

AFTER:
improvement

galgeek:
Thank you, @vbanos!

We'll get set up for a test crawl early next week.

PR #257, opened 2023-07-07
Do not try to get a screenshot if status is 4xx, 5xx

vbanos:
The screenshot is an additional thing we do when the capture is successful. Why get a screenshot of 4xx/5xx responses? Its just extra system load.
We already got the capture for archiving reasons.


galgeek:
Is it reasonable to use this code only if `simpler404` is set?

vbanos:
@galgeek good point. PR updated. Thanks.

PR #270, opened 2024-02-02
MOMA behavior

galgeek:
Capture more content from current https://www.moma.org/ artist pages


galgeek:
note: approved separately by @avdempsey

PR #271, opened 2024-02-08
Use black, enforce with GitHub Actions

avdempsey:
The Makefile is configured to run black to format with support for py35 - py312.


galgeek:
looks good to me — thank you, @avdempsey!

PR #276, opened 2024-04-17
run yt-dlp after brozzling a page (if at all)

galgeek:
brozzler should capture pages before video

### DIFF
--- a/brozzler/worker.py
+++ b/brozzler/worker.py
@@ -258,7 +258,7 @@ class BrozzlerWorker:
             except brozzler.PageInterstitialShown:
                 self.logger.info("page interstitial shown (http auth): %s", page)
 
-            if enable_youtube_dl and ydl.should_ytdlp(page):
+            if enable_youtube_dl and ydl.should_ytdlp(page, site):
                 try:
                     ydl_outlinks = ydl.do_youtube_dl(self, site, page)
                     outlinks.update(ydl_outlinks)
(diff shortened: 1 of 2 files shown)


brozzler/ydl.py
```diff
@@ -32,6 +33,30 @@
 thread_local = threading.local()
 
 
+def is_html_maybe(url):
+    skip_url_exts = ["pdf", "jpg", "jpeg", "png", "gif", "mp4", "mpeg"]
```
avdempsey, line 37: .css and .js might be good too

brozzler/ydl.py
```diff
@@ -32,6 +33,33 @@
 thread_local = threading.local()
 
 
+def is_html_maybe(url):
+    if "chrome-error:" in url:
+        return False
+
+    skip_url_exts = ["pdf", "jpg", "jpeg", "png", "gif", "mp3", "mp4", "mpeg", "css", "js"]
```
vbanos, line 40: Use `set([...])` for faster lookups.

vbanos, line 40: This list of extensions is really limited. I suggest that you do something like the following outside the function and then use `EXTENSIONS` inside it.
```
import mimetypes
EXTENSIONS = list(mimetypes.types_map) + list(mimetypes.common_types)
EXTENSIONS.remove('.htm')
EXTENSIONS.remove('.html')
```

vbanos:
Improvement suggestion: `browse_page` also gets the HTTP status of the response. You can access it via `browser.websock_thread.page_status`. If its not `200`, you shouldn't run yt-dlp.

vbanos:
That's all from me, I think that this is a good improvement.

brozzler/worker.py
```diff
@@ -242,53 +242,98 @@ def brozzle_page(
...
+                last_modified_header = "Last-Modified"
+            elif "Last-Modified" in r.headers:
+                last_modified_header = "Last-Modified"
+            elif "LAST-MODIFIED" in r.headers:
+                last_modified_header = "LAST-MODIFIED"
+            if last_modified_header:
+                page.last_modified = r.headers[last_modified_header]
+                self.logger.info(
+                    "url %s last_modified is %s", page.url, page.last_modified
+                )
```
vbanos, line 330: I have a suggestion to make a code improvement here.
If you use a dict-like data structure that is **case insensitive** like this one
https://github.com/psf/requests/blob/main/src/requests/structures.py#L13
you won't have to write all this code and you'll catch ALL cases.
E.g. maybe you are missing some edge cases like `LAST-modified`, people do weird things on the web :) 
I mention this because you already have `requests` in your dependencies and it would be easy to use it.
An alternative one is this one : https://werkzeug.palletsprojects.com/en/3.0.x/datastructures/#werkzeug.datastructures.Headers

vbanos: Oh, I'm sorry but now that I'm looking it again, I see that you are checking a response from the `request` libraries. Maybe the data structure I mentioned is already used and you can take advantage of it right now. Please try to remove all these alternative upper-case / lower-case combinations and try to use only lower-case to check that it works.

galgeek: yes, thank you, @vbanos!  `requests` uses its `CaseInsensitiveDict` for the headers here, and I've updated.

brozzler/worker.py
```diff
@@ -242,53 +242,80 @@ def brozzle_page(
...
+                    else:
+                        self.logger.error(
+                            "youtube_dl raised exception on %s", page, exc_info=True
+                        )
+        else:
+            self.logger.info("needs fetch: %s", page)
+            self._fetch_url(site, page=page)
         return outlinks
 
+    def _get_page_headers(self, page):
```
adam-miller, line 293: Maybe worth a comment here to note that we are doing an abbreviated request and bypassing warcprox, thus avoiding a partial capture warc record.

brozzler/worker.py
```diff
@@ -242,53 +242,80 @@ def brozzle_page(
...
+                            "youtube-dl got %s %s processing %s",
+                            e.exc_info[1].code,
+                            e.exc_info[1].msg,
+                            page.url,
+                        )
+                    else:
+                        self.logger.error(
+                            "youtube_dl raised exception on %s", page, exc_info=True
+                        )
+        else:
```
adam-miller, line 288: Style nitpick, but it can be cleaner if this else block is handled before the main case by doing something like this around line 249:
```
if not self._needs_browsing(page):
    self.logger.info("needs fetch: %s", page)
    self._fetch_url(site, page=page)
    return outlinks
```

adam-miller (approved):

PR #277, opened 2024-04-25
skip ytdlp for selected seeds

galgeek:
brozzler should skip running yt-dlp for selected seeds


brozzler/cli.py
```diff
@@ -544,12 +544,25 @@ def dump_state(signum, frame):
         finally:
             signal.signal(signal.SIGQUIT, dump_state)
 
+    def get_skip_av_seeds():
+        skip_av_seeds_file = "/opt/local/brozzler/skip_av_seeds.txt"
```
gretchenleighmiller, line 548: nit: This is a constant, so it should be upper cased.

brozzler/cli.py
```diff
@@ -544,12 +544,25 @@ def dump_state(signum, frame):
...
+        try:
+            with open(skip_av_seeds_file) as skips:
+                skip_av_seeds = {x for x in skips.readlines()}
```
gretchenleighmiller, line 551: Could you explain what the expected data structure is for this value? You're using `readlines` on a text file, which should return a list of lines, but you're trying to using a dictionary comprehension on the resulting list.

This would be clearer if you defined the expected input file format and the expected structure that should result in this variable. (Leaving the details of that up to you, but would be happy to assist.)

brozzler/ydl.py
```diff
@@ -32,9 +32,20 @@
...
+        return False
+    if site.skip_ytdlp:
+        logging.info("skipping ytdlp: site marked skip_ytdlp")
+        return False
+
+    ytdlp_seed = site["metadata"]["ait_seed_id"] if "metadata" in site and "ait_seed_id" in site["metadata"] else None
+
+    if ytdlp_seed and ytdlp_seed in worker.skip_av_seeds:
+        logging.info("skipping ytdlp: site in skip_av_seeds")
+        site.skip_ytdlp = True
```
adam-miller, line 48: Maybe we should also cache the False condition as well, so we avoid the scan of the `skip_av_seeds` list
Then we'd modify the conditional check around line 40 to handle it being True/False or None

brozzler/ydl.py
```diff
@@ -32,9 +32,20 @@
 thread_local = threading.local()
 
 
-def should_ytdlp(page, site):
+def should_ytdlp(worker, site, page):
     # called only after we've passed needs_browsing() check
     if page.status_code != 200:
+        logging.info("skipping ytdlp: non-200 page status")
+        return False
+    if site.skip_ytdlp:
```
adam-miller, line 40: I think we need to add the `skip_ytdlp` attribute to the Site class in `model.py`

adam-miller: After some discussion, it turns out this is handled for us in the doublethink library here: https://github.com/internetarchive/doublethink/blob/1dc2e0395ca697115a38e562e81eb922577bea41/doublethink/orm.py#L124C5-L125C1

brozzler/cli.py
gretchenleighmiller: OK, I checked this, and I see that it's a set comprehension. You could just skip the comprehension and cast this to a set: `set(skips.readlines())`.

brozzler/ydl.py
galgeek: nevertheless, I've added to `Site`'s `populate_defaults`

brozzler/cli.py
galgeek: thanks, updated

brozzler/cli.py
galgeek: thanks, updated

brozzler/ydl.py
galgeek: I prefer simpler code with default = False, I think

gretchenleighmiller (approved):

adam-miller (approved):

adam-miller (approved):

gretchenleighmiller (approved):

PR #279, opened 2024-06-17
Update Ansible and Vagrant for Debian 12

insom:
Ubuntu no longer packages `chromium` as a .deb, only as a snap. Debian is keeping the .deb dream alive.

I've made the playbook more resilient to path changes with Python versions by using `-mvirtualenv` instead of executing the module file and using `python3` instead of expecting `python` on the system path to point to the right place.

This proposed PR would lose the ability to archive Adobe Flash as no equivalent is packaged (or would work with a modern Chromium, I think?)

### DIFF
--- a/ansible/roles/brozzler-worker/tasks/main.yml
+++ b/ansible/roles/brozzler-worker/tasks/main.yml
@@ -51,7 +51,7 @@
   file: path={{venv_root}}/websockify-ve3 state=directory owner={{user}}
  
 #get python3 version for checks below 
-- shell: "python --version"
+- shell: "python3 --version"
   register: python_installed
   
  #websockify's dependency numpy's latest version no longer supports 3.5
@@ -60,7 +60,7 @@
     name: numpy==1.18.4
     virtualenv: '{{venv_root}}/websockify-ve3'
     virtualenv_python: python3
-    virtualenv_command: python3 /usr/lib/python3/dist-packages/virtualenv.py
+    virtualenv_command: python3 -mvirtualenv
     extra_args: '--no-input --upgrade --pre --cache-dir=/tmp/pip-cache'
   become: true
   become_user: '{{user}}'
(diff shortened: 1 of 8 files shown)


ansible/roles/brozzler-worker/tasks/main.yml
```diff
@@ -1,24 +1,15 @@
...
-
 - name: ensure required packages are installed
   become: true
   apt: name={{item}} state=present
   with_items:
-  - chromium-browser
-  - vnc4server
-  - libjpeg-turbo8-dev
+  - chromium
+  - tigervnc-standalone-server
```
insom, line 7: `vnc4server` has a proprietary license so I switched over to `tigervnc` which is in the main repositories.

ansible/roles/common/tasks/main.yml
```diff
@@ -70,7 +70,7 @@
   changed_when: false
 
 - name: ensure service user {{user}} exists
-  user: name={{user}} system=yes createhome=no home=/nonexistent
+  user: name={{user}} system=yes createhome=yes home=/home/{{user}}
```
insom, line 73: This user does actually have to exist, as `{{user}}` is used later in the `warcproxy` playbook. The only reason this didn't error before is because `/home/vagrant` is created by default by Vagrant itself. (I was running inside Vagrant though, so I hit this).

ansible/roles/common/tasks/main.yml
```diff
@@ -70,7 +70,7 @@
...
         shell=/usr/sbin/nologin
   become: true
-  when: id_user|failed
+  when: id_user.rc != 0
```
insom, line 76: This is as suggested by the Ansible docs, although #218 suggests `is failed` which is nicer to read? `|failed` filter definitely does not work any longer.

insom:
(converting to draft because it looks like Xvnc is crash looping in this configuration)

PR #281, opened 2024-07-28
Upgrade websocket-client dependency

vbanos:
We use a very old version (0.47.0 was released on Feb 22, 2018). This project is actively developed so it would be great to upgrade. Changelog: https://github.com/websocket-client/websocket-client/blob/master/ChangeLog

I tested every version from 0.48.0 to 0.59.0.
Only versions 0.52.0, 0.58.0 and 0.59.0 worked correctly. All others raised exceptions. Thus, I suggest to use 0.59.0 for now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ setuptools.setup(
         "yt_dlp<2023.11.16",
         "reppy==0.3.4",
         "requests>=2.21",
-        "websocket-client>=0.39.0,<=0.48.0",
+        "websocket-client==0.59.0",
         "pillow>=5.2.0",
         "urlcanon>=0.1.dev23",
         "doublethink @ git+https://github.com/internetarchive/doublethink.git@Py311",


vbanos:
Furthermore, I tested version 1.0.0 to 1.4.3. Captures worked but in the end, all raised the following exception:
```                                                                             
[2024-07-28 10:51:06,819: ERROR/ForkPoolWorker-2] error from callback <bound method WebsockReceiverThread._on_close of <WebsockReceiverThread(WebsockThread:47001, started daemon 139974406829824)>>: _on_close() takes 2 positional arguments but 4 were given
[2024-07-28 10:51:06,819: ERROR/ForkPoolWorker-2] exception from websocket receiver thread
  File "/opt/spn/lib/python3.8/site-packages/websocket/_app.py", line 407, in _callback
    callback(self, *args)                                                       
TypeError: _on_close() takes 2 positional arguments but 4 were given            
```                                                                             
                                                                                
We need to change Brozzler code to use version 1+.

So, I recommend to upgrade to 0.59.0 as a first step and then later make the necessary code changes to upgrade even more.
Thank you.

vbanos:
Also, the new websocket client will be needed to use the latest Chrome.

vbanos:
Already done at https://github.com/internetarchive/brozzler/commit/b41393fac5353fe01154c90fc54ee532e149ec59

PR #284, opened 2024-08-12
fix: handle exceptions when requesting page headers for content-type …

adam-miller:
…sampling.

### DIFF
--- a/brozzler/worker.py
+++ b/brozzler/worker.py
@@ -295,9 +295,13 @@ class BrozzlerWorker:
     def _get_page_headers(self, page):
         # bypassing warcprox, requests' stream=True defers downloading the body of the response
         # see https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow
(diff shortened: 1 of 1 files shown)


brozzler/worker.py
```diff
@@ -295,9 +295,13 @@ def brozzle_page(
     def _get_page_headers(self, page):
         # bypassing warcprox, requests' stream=True defers downloading the body of the response
         # see https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow
-        with requests.get(page.url, stream=True) as r:
-            page_headers = r.headers
-        return page_headers
+        try:
+            with requests.get(page.url, stream=True, verify=False) as r:
+                page_headers = r.headers
+                return page_headers
```
galgeek, line 301: it might be good to preserve the original indentation here

adam-miller: Fixed. Thanks!

galgeek (approved):
looks good to me — thank you!

PR #285, opened 2024-08-30
feat: Detect connection failures forwarded from warcprox and retry th…

adam-miller:
…em with backoff


adam-miller:
Brozzler didn't really have a retry loop for failed pages except when connection to warcprox failed, and this too would fail in a tight loop. Warcprox connection failures and timeouts are returned to the browser as 502 and 504 status codes, so I'm checking for those and adding a retry loop with backoff. This is accomplished by adding a `retry_after` field to the page in rethinkdb, and then adjusting the query for claiming a page. This then causes a tight loop on claiming the site, so I add a delay there to avoid attempting to immediately claim a site that was just disclaimed.

brozzler/worker.py
```diff
@@ -499,11 +502,23 @@ def brozzle_site(self, browser, site):
...
+                    site,
+                    page,
+                    exc_info=True,
+                )
+            else:
+                self.logger.error(
+                    "unexpected exception site=%r page=%r", site, page, exc_info=True
+                )
             if page:
+                retry_delay = min(60, 60 * (1.5**page.failed_attempts))
```
adam-miller, line 518: This `min()` should definitely be `max()` and also needs a comment explaining that this is a simple backoff calculation yielding 60, 90, 135 for attempts of 0, 1, 2

brozzler/worker.py
```diff
@@ -254,10 +255,12 @@ def brozzle_page(
         else:
             self.logger.info("needs browsing: %s", page)
             try:
-                browser_outlinks = self._browse_page(
+                browser_outlinks, status_code = self._browse_page(
```
adam-miller, line 258: Maybe skip returning this here and use `browser.websock_thread.page_status`

brozzler/worker.py
```diff
@@ -499,11 +502,23 @@ def brozzle_site(self, browser, site):
...
-            )
+        except (brozzler.PageConnectionError, Exception) as e:
+            if isinstance(e, brozzler.PageConnectionError):
+                self.logger.error(
+                    "Page status code possibly indicates connection failure between host and warcprox: site=%r page=%r",
+                    site,
+                    page,
+                    exc_info=True,
+                )
+            else:
```
adam-miller, line 513: Better to leave the default except block alone, and just have an except brozzler.PageConnectionError block above it

brozzler/frontier.py
```diff
@@ -138,7 +138,13 @@ def claim_sites(self, n=1):
                         emit=lambda acc, site, new_acc: r.branch(
                             r.and_(
                                 r.or_(
-                                    site["claimed"].not_(),
+                                    r.and_(
+                                        site["claimed"].not_(),
+                                        r.or_(
+                                            site.has_fields("last_disclaimed").not_(),
+                                            site["last_disclaimed"].lt(r.now().sub(20)),
```
adam-miller, line 145: Maybe a better way to do this, but at least comment the intent. We are ensuring a site cannot be reclaimed within 20s of it having been disclaimed to avoid the tight retry loop when the only remaining pages have a retry_after set in the future.

brozzler/worker.py
```diff
@@ -499,11 +502,23 @@ def brozzle_site(self, browser, site):
...
+                self.logger.error(
+                    "unexpected exception site=%r page=%r", site, page, exc_info=True
+                )
             if page:
+                retry_delay = min(60, 60 * (1.5**page.failed_attempts))
+                page.retry_after = doublethink.utcnow() + datetime.timedelta(
+                    seconds=retry_delay
+                )
                 page.failed_attempts = (page.failed_attempts or 0) + 1
```
adam-miller, line 522: If we're setting this in populate_defaults we don't need to have a default here

brozzler/frontier.py
gretchenleighmiller: I would suggest assigning this to a constant to at least make it semantically clearer. Or maybe put a little more detail in the comment: "We are ensuring a site cannot be reclaimed within 20s of it having been disclaimed to avoid the tight retry loop when the only remaining pages have a retry_after set in the future." would be sufficient.

brozzler/worker.py
```diff
@@ -535,11 +539,23 @@ def brozzle_site(self, browser, site):
...
+                    site,
+                    page,
+                    exc_info=True,
+                )
+            else:
+                self.logger.error(
+                    "unexpected exception site=%r page=%r", site, page, exc_info=True
+                )
             if page:
+                retry_delay = min(135, 60 * (1.5**page.failed_attempts))
```
gretchenleighmiller, line 555: I would also suggest making this slightly clearer with a comment.

gretchenleighmiller:
I'm not totally sure if you've finished working on this yet, but mostly lgtm. Just suggesting some additional clarity around the magic numbers.

brozzler/worker.py
```diff
@@ -549,7 +567,8 @@ def brozzle_site(self, browser, site):
                         page,
                     )
                     self._frontier.completed_page(site, page)
-                    page = None
```
galgeek, line 552: do we really want to skip this line?

adam-miller: Good catch. You're correct that this seems necessary in disclaim_site() when we're done retrying a page.

PR #287, opened 2024-09-11
prometheus metrics for brozzler, plus yt-dlp

galgeek:


brozzler/cli.py
```diff
@@ -234,6 +234,24 @@ def brozzle_page(argv=None):
         action="store_true",
         help="Try to avoid web bot detection",
     )
+    arg_parser.add_argument(
```
wilsoniya, line 237: consider adding: 

```
    type=int,
```

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+    from http_sd_registry.client import (
+        Client,
+        Env,
+        Registration,
+        Scheme,
+        format_self_target,
+    )
+    from http_sd_registry.config import ClientConfig
+except ImportError:
+    http_sd_registry = None
```
wilsoniya, line 13: Why would there be an `ImportError`? Is it because FOSS users may not have access to non-FOSS `http_sd_registry`? Consider explaining the rationale (regardless of what it is) as a comment here.

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+from prometheus_client import Counter, Gauge, Histogram, start_http_server
+
+# fmt: off
+brozzler_pages_crawled = Counter("brozzler_pages_crawled", "number of pages visited by brozzler")
+brozzler_page_processing_duration_seconds = Histogram("brozzler_page_processing_duration_seconds", "time spent processing a page in brozzler")
```
wilsoniya, line 20: The default Histogram buckets are likely not appropriate for this use case; the maximum bucket is `10`: https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics.py#L586

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+brozzler_outlinks_found = Counter("brozzler_outlinks_found", "number of outlinks found by brozzler")
+brozzler_last_page_crawled_time = Gauge("brozzler_last_page_crawled_time", "time of last page visit")
```
wilsoniya, line 22: consider elaborating on the units in use here, e.g.: 

```suggestion
brozzler_last_page_crawled_time = Gauge("brozzler_last_page_crawled_time", "time of last page visit expressed in seconds since the UNIX epoch")
```

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+brozzler_in_progress_pages = Gauge("brozzler_in_progress_pages", "number of pages currently processing with brozzler")
+brozzler_resources_requested = Counter("brozzler_resources_requested", "number of resources requested", labelnames=["resource_type"])
```
wilsoniya, line 24: How many different `resource_type` label values are we expecting? Each unique label value combination produces another series, so we want to make sure label value cardinality is bounded and not super huge.

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+brozzler_resources_fetched = Counter("brozzler_resources_fetched", "number of resources fetched", labelnames=["resource_type", "status_code"])
+brozzler_resources_size_total = Counter("brozzler_resources_size_total", "total size of resources fetched", labelnames=["resource_type"])
+brozzler_resources_fetch_time = Counter("brozzler_resources_fetch_time", "time spent fetching resources", labelnames=["resource_type"])
+brozzler_ydl_urls_checked = Counter("brozzler_ydl_urls_checked", "count of urls checked by brozzler yt-dlp")
+brozzler_ydl_download_attempts = Counter("brozzler_ydl_download_attempts", "count of download attempted by brozzler yt-dlp", labelnames=["host"])
```
wilsoniya, line 29: What's the meaning of the `host` label? If it's the host on which brozzler is running, this label is very likely applied automatically by prometheus.

brozzler/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+        env_for_prom = Env.qa
+    elif env == "prod":
+        env_for_prom = Env.prod
+
+    config = ClientConfig(server_url_base=registry_url)
+    client = Client(config)
+    target = format_self_target(scrape_port=metrics_port)
+    registration = Registration(
+        target=target,
+        env=env_for_prom,
```
wilsoniya, line 56: `env` is required in the `Registration` type, and may not be `None`, however your function declares `env` as optional. Strongly consider requiring the parameter, and/or raising fi the value is provided as `None`. 

Even though your CLI may require a value for `env`, 
1. currently the CLI doesn't assert that a valid `Env` value is passed, which may cause this function to define `env` as `None`
2. a future brozzler author may reuse `register_prom_metrics`, and we should make it difficult (ideally impossible) to provide invalid parameters.

galgeek:
Note: I'm focusing now on updating #289, which replaces this, and plan to close this PR soon.

galgeek:
closing in favor of #289

PR #288, opened 2024-09-12
Implement Seed-level video capture setting handling + Job-level PDF-only option

gretchenleighmiller:
This PR covers the following:
- Adds a new `video_capture` configuration option on the Seed level. This has four possible values; see newly added documentation for details.
- Implements the `video_capture` option, which impacts `yt-dlp` extraction and MIME types of outlinks. The remainder of video capture handling is accomplished in warcprox via `Warcprox-Meta` headers.
- Removes the previous file-based `skip_av_seeds` functionality.
- Adds a new `pdfs_only` configuration option on the Job level. This is a boolean that defaults to `False`; see newly added documentation for details.
- Implements the `pdfs_only` option based on the MIME type of outlinks. The remaining PDF-only filtering is accomplished in warcprox via `Warcprox-Meta` headers.
- Documentation updates and minor style cleanup.


brozzler/worker.py
```diff
@@ -250,7 +249,17 @@ def brozzle_page(
 
         if not self._needs_browsing(page_headers):
             self.logger.info("needs fetch: %s", page)
-            self._fetch_url(site, page=page)
+            if site.video_capture in [
```
galgeek, line 252: with if and elif, shouldn't pdfs_only should get checked first?

job-conf.rst
```diff
@@ -107,6 +107,18 @@ Puts a cap on the number of sites belonging to a given job that can be brozzled
...
 seeds starving out other jobs.
 
+``pdfs_only``
+~~~~~~~~~~~~~~~~~~~~~
++---------+----------+-----------+
+| type    | required | default   |
++=========+==========+===========+
+| boolean | no       | ``false`` |
++---------+----------+-----------+
+Limits capture to PDFs based on MIME type. This value will only impact
```
galgeek, line 117: Current code changes only processing of page-level urls with PDF-ish content-type.

job-conf.rst
```diff
@@ -107,6 +107,18 @@ Puts a cap on the number of sites belonging to a given job that can be brozzled
...
+processing of outlinks within Brozzler. Fully limiting a crawl to only PDFs
+requires an additional entry in the Warcprox-Meta header ``mime-type-filters``
+key to fully block videos by MIME type.
```
galgeek, line 120: should "block videos" be "block PDFs" here?

job-conf.rst
```diff
@@ -107,6 +107,18 @@ Puts a cap on the number of sites belonging to a given job that can be brozzled
 simultaneously across the cluster. Addresses the problem of a job with many
 seeds starving out other jobs.
 
+``pdfs_only``
```
galgeek, line 110: Finishing up the warcprox-meta implementation here could simplify updating this documentation.

brozzler/worker.py
```diff
@@ -303,13 +312,29 @@ def _get_page_headers(self, page):
...
+        Determines if the page's Content-Type header specifies that it contains
+        a video.
+        """
+        return (
+            "content-type" in page_headers and "video" in page_headers["content-type"]
+        )
+
+    def _is_pdf(self, page_headers) -> bool:
+        """
+        Determinse if the page's Content-Type header specifies that it is a PDF.
```
galgeek, line 332: fix "Determinse" typo

galgeek (changes requested):
@gretchenleighmiller, thanks for your work on this!

I've left a couple of comments it might be good to address.

brozzler/worker.py
gretchenleighmiller: It should have the correct behavior either way.

The first conditional is checking that 1) there is a rule to limit video capture by MIME type AND 2) that the MIME type specified in the Content-Type header of the response indicates that it's a video. Both have to be true to enter that branch, so we aren't going to accidentally skip the PDF conditional if Content-Type is a PDF.

The second conditional is similarly constrained on the PDF-only option being set and the Content-Type header specifying that the response is a PDF.

I would not expect the order to matter, but maybe I'm missing something?

job-conf.rst
gretchenleighmiller: I'm not sure what you're suggesting. I think we need the callout in here that this option should be paired with a MIME type filter defined as part of `warcprox_meta`. Do you think there's a clearer way to document that in here?

job-conf.rst
gretchenleighmiller: Are you suggesting a better way of wording this in the documentation? It is called out here that this option specifically has limited impact and must be paired with a `warcprox_meta` rule to further filter by MIME type.

Do you think "PDF-ish" is better phrasing given the ambiguity of MIME type in Content-Type headers? I would expect MIME type to be pretty straightforward in this case given that PDFs are a pretty well-defined and well-known format.

brozzler/worker.py
gretchenleighmiller: Fixed, thank you!

job-conf.rst
gretchenleighmiller: That clause in the sentence was redundant, so I just deleted it. Thank you!

brozzler/worker.py
galgeek: thanks! I did miss `self._is_video_type(page_headers)` initially.

I still like starting with the pdfs_only check — it seems simpler, and encompasses any limiting VideoCaptureOptions enabled.

job-conf.rst
galgeek: Something like this would be better, I think:

"Limits captures to PDFs, based on page's content-type header.
This value, for now, affects only brozzler's capture of page-level urls. 

Note: fully limiting a crawl to PDFs only will also require updates to brozzler's Warcprox-Meta header and warcprox."

brozzler/worker.py
gretchenleighmiller: Swapped them.

job-conf.rst
gretchenleighmiller: Reworded slightly for clarity.

brozzler/cli.py
```diff
@@ -598,27 +598,12 @@ def dump_state(signum, frame):
         finally:
             signal.signal(signal.SIGQUIT, dump_state)
 
-    def get_skip_av_seeds():
```
galgeek, line 601: is it important to remove this?

it's been good to have the ability to skip yt-dlp for seeds administratively. maybe this is also helpful to others using brozzler?

gretchenleighmiller: I don't think there are any other Brozzler users leveraging this functionality. It's not documented anywhere and requires a text file existing at a hard-coded location.

The correct way to do this is to move the functionality over into job config and document it, which is what this PR does. We should EOL usage of this feature internally and move config for the affected seeds over so they continue to have AV capture skipped.

avdempsey: Sometimes we discover a problematic seed while a crawl job is running. If we remove `get_skip_av_seeds` it would be good to have a procedure we can use to unstick the cluster.

galgeek (approved):

brozzler/cli.py
gretchenleighmiller: Can we have a process to remove/update configuration of problematic seeds directly in the job config in RethinkDB?

This implementation is pretty messy to maintain alongside the job config updates to support video exclusion, in particular because it gets passed down into should_ytdlp in brozzler/ydl.py. I don't want us to have to maintain two processes that do functionally the same thing. Ideally we can drive everything from job config, including in-place updates.

brozzler/worker.py
```diff
@@ -399,13 +395,29 @@ def _get_page_headers(self, site, page):
             url_logger.warning("Failed to get headers", exc_info=True)
         return {}
 
-    def _needs_browsing(self, page_headers):
-        if (
+    def _needs_browsing(self, page_headers) -> bool:
+        return not bool(
```
mistydemeo, line 399: Does this need to be cast to a `bool`?

mistydemeo (approved):
Right now it looks like `DISABLE_VIDEO_CAPTURE` and `DISABLE_YTDLP_CAPTURE` are always doing the same thing; is there a plan tohave them branch at some point?

brozzler/worker.py
gretchenleighmiller: Good catch! I think I originally had that as belt-and-suspenders, but I'm not doing that on the other similar methods, so might as well get rid of it.

brozzler/worker.py
```diff
@@ -274,36 +273,23 @@ def thumb_jpeg(self, full_jpeg):
...
             logger.info("skipping ytdlp: non-200 page status", page_status=page_status)
             return False
-        if site.skip_ytdlp:
-            logger.info("skipping ytdlp: site marked skip_ytdlp")
-            return False
+        if site.video_capture in [
+            VideoCaptureOptions.DISABLE_VIDEO_CAPTURE.value,
+            VideoCaptureOptions.DISABLE_YTDLP_CAPTURE.value,
+        ]:
+            logger.info("skipping ytdlp: site has video capture disabled")
```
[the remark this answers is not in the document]
gretchenleighmiller: Ahhhhh, you're right! This was a merge conflict resolution mistake. Fixed, thank you!

PR #294, opened 2024-10-17
Add support for Chrome > 110

vbanos:


vbanos:
Tested with Google Chrome `130.0.6723.58-1` and `110.0.5481.177-1`. Both returned the same results in my experiments.

Chrome 130 seemed ~25% slower but I didn't do a large number of experiments to be totally sure about this.

galgeek (approved):

galgeek:
Thank you, @vbanos!

PR #303, opened 2024-11-12
Disable automatic http to https transformation

vbanos:
Chrome 130 automatically converts http to https even if the target URL is http. We disable this behavior because some target sites simply don't have https.


vbanos:
Relevant: https://github.com/puppeteer/puppeteer/issues/10791

PR #307, opened 2024-12-07
brozzler yt-dlp should be able to specify a separate tempdir

galgeek:
adds a command-line param to specify a separate tmp dir for yt-dlp use — defaults to /tmp


brozzler/ydl.py
```diff
@@ -420,7 +420,7 @@ def do_youtube_dl(worker, site, page):
     Returns:
          `list` of `str`: outlink urls
     """
-    with tempfile.TemporaryDirectory(prefix="brzl-ydl-") as tempdir:
+    with tempfile.TemporaryDirectory(prefix="brzl-ydl-", dir="/tmp") as tempdir:
```
gretchenleighmiller, line 423: Consider making this configurable instead of adding a new magic string.

gretchenleighmiller (changes requested):

adam-miller (approved):

PR #308, opened 2024-12-07
retry only proxied urls

galgeek:


brozzler/ydl.py
```diff
@@ -285,6 +285,7 @@ def ydl_postprocess_hook(d):
     ytdlp_url = page.redirect_url if page.redirect_url else page.url
     is_youtube_host = isyoutubehost(ytdlp_url)
     if is_youtube_host and YTDLP_PROXY:
+        MAX_YTDLP_ATTEMPTS = 4
```
gretchenleighmiller, line 288: Redefining this here is very likely to be problematic. While Python doesn't strictly enforce this, `MAX_YTDLP_ATTEMPTS` is, semantically speaking, a constant. We shouldn't be redefining this again inline in a function. But furthermore, redefining this here for one URL being processed will redefine it for _every_ URL processed for the lifetime of the process because `MAX_YTDLP_ATTEMPTS` is defined in module scope.

gretchenleighmiller (changes requested):

brozzler/ydl.py
```diff
@@ -33,9 +33,11 @@
...
 thread_local = threading.local()
 
-YTDLP_PROXY = ""
-MAX_YTDLP_ATTEMPTS = 4
-YTDLP_WAIT = 10
+
+ytdlp_proxy = ""
+ytdlp_tmp = "/tmp"
+ytdlp_wait = 10
+max_ytdlp_attempts = 1
```
gretchenleighmiller, line 40: Changing these to lowercase doesn't fix the problem. You're still redefining what should be a constant; it's just harder to tell that it should be a constant because you made it lowercase.

`max_ytdlp_attempts` is module-level scope. So when you redefine `max_ytdlp_attempts` from `1` to `4` inside of `should_ytdlp`, it does not revert back to `1` after exiting the function. It will remain as `4`, which breaks the entire point of what you are trying to do.

gretchenleighmiller (changes requested):

adam-miller (approved):

PR #312, opened 2025-01-03
feat: override yt-dlp generic extractor to add redirect loop detectio…

adam-miller:
…n logic


adam-miller:
This is a workaround for this yt-dlp bug: https://github.com/yt-dlp/yt-dlp/issues/4752

galgeek (approved):

PR #315, opened 2025-01-23
yt-dlp proxy handling update

galgeek:


gretchenleighmiller (changes requested):
I would _very strongly_ prefer that we do not continue the pattern of adding configuration options through these hooks that only accept a text file at a hardcoded location.

There is an established pattern for adding configuration options in Brozzler through additional command-line options. That is where yt-dlp proxying configuration should be added. It could be through proxies defined directly on the command line or pointing to a file containing the proxy definition.

adam-miller (approved):

PR #320, opened 2025-02-07
feat: add chrome flag to disable download of autocomplete optimizatio…

adam-miller:
…n models at launch


galgeek (approved):

vbanos:
We already have an existing `--disable-features=HttpsUpgrades`  option, add your extra flags to that instead of adding another one.

galgeek (approved):

PR #322, opened 2025-02-11
requests timeout for fetch_url, plus user_agent

galgeek:


brozzler/worker.py
```diff
@@ -480,13 +488,22 @@ def _fetch_url(self, site, url=None, page=None):
                 "http": "http://%s" % self._proxy_for(site),
                 "https": "http://%s" % self._proxy_for(site),
             }
+        user_agent = site.get("user_agent")
+        headers = {"User-Agent": user_agent} if user_agent else {}
+        headers.update(site.extra_headers(page))
```
adam-miller, line 493: I'm surprised we weren't doing this already. We may want to use a library that more closely resembles a browser for `_fetch_url()` in the future, but this is a good step.

brozzler/worker.py
```diff
@@ -330,12 +331,19 @@ def brozzle_page(
...
+            user_agent = site.get("user_agent")
+            headers = {"User-Agent": user_agent} if user_agent else {}
+            self.logger.info("getting page headers for %s", page.url)
             with requests.get(
-                page.url, stream=True, verify=False, timeout=self.HEADER_REQUEST_TIMEOUT
+                page.url,
+                stream=True,
+                verify=False,
+                headers=headers,
+                timeout=self.HEADER_REQUEST_TIMEOUT,
```
adam-miller, line 346: Indeed, we want to avoid a default requests lib header for something like this. We're more likely to get results if our request looks more normal.

adam-miller (approved):

adam-miller (approved):

PR #323, opened 2025-02-13
better error handling for _fetch_url

galgeek:


brozzler/worker.py
```diff
@@ -480,34 +482,39 @@ def _on_service_worker_version_updated(chrome_msg):
...
+
         user_agent = site.get("user_agent")
         headers = {"User-Agent": user_agent} if user_agent else {}
         headers.update(site.extra_headers(page))
 
         self.logger.info("fetching url %s", url)
         try:
             # response is ignored
-            requests.get(
+            http.request(
```
adam-miller, line 502: It would be nice if we had an equivalent of max fetch time, but I don't see anything available out of the box, so it can wait.

PR #324, opened 2025-02-19
CLI: improve Chrome finding on Mac

mistydemeo:
On macOS, we can find Chrome even if it's installed in a non-default path by querying `mdfind`. This is the CLI entrypoint to Spotlight, and we can use it to look up applications using their unique bundle identifiers.

If `mdfind` fails to find anything, this falls back to the hardcoded paths. This should ensure this still works if Spotlight indexing is off, but Chrome is in the default path. If we'd prefer, I can switch this to looking up the hardcoded paths first and fall back to `mdfind` if that doesn't work instead.

### DIFF
--- a/brozzler/cli.py
+++ b/brozzler/cli.py
@@ -29,6 +29,7 @@ import requests
 import doublethink
 import signal
 import string
+import subprocess
 import sys
 import threading
 import time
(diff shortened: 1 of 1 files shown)


brozzler/cli.py
```diff
@@ -126,8 +127,38 @@ def configure_logging(args):
...
+
 def suggest_default_chrome_exe():
-    # mac os x application executable paths
+    # First ask mdfind, which lets us find it in non-default paths
+    if sys.platform == "darwin":
+        path = suggest_default_chrome_exe_mac()
+        if path is not None and os.path.exists(path):
+            return path
+
+    # Fall back to default paths if mdfind couldn't find it
```
galgeek, line 161: maybe the block beginning here should live in `suggest_default_chrome_exe_mac` too?

mistydemeo: Good call; moved all of that in.

galgeek (approved):

PR #325, opened 2025-02-20
Set up structlog logging

mistydemeo:
This ports the logging from `logging` to `structlog`. I've updated all of the logger instantiations along with all of the places `logging` was called. I've broken out most data that was being inlined into log statements so that it's now structured arguments to the log statements instead. I've also updated at least a few places to add some default parameters that will be included in all logging statements, but I think it'd be good to look at what other common variables we could use to link multiple log statements together. Specifically, at the moment:

* The yt-dlp class always logs the URL
* BrozzlerWorker always logs the thread

To start, I've kept structlog's default formatting. I know we're interested in keeping text-based logs as the default for now; we can talk whatever specifics we'd like. Here's an example of what it looks like at the moment.

```
2025-02-21 09:09:46 [info     ] adding ssurt to scope accept rules [brozzler.model.Site._accept_ssurt_if_not_redundant(model.py:284)] ssurt=com,example,//https:/
2025-02-21 09:09:46 [warning  ] not starting prometheus scrape endpoint: metrics_port is undefined [brozzler.worker.BrozzlerWorker.__init__(worker.py:129)]
2025-02-21 09:09:46 [info     ] running                        [brozzler.chrome.Chrome.start(chrome.py:224)] chrome_args='"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" -v --remote-debugging-port=9222 --remote-allow-origins=http://localhost:9222 --use-mock-keychain --user-data-dir=/var/folders/v9/fhm3y0l12zvbmtpmf39nxm6h0000gn/T/tmp9g9tmai0/chrome-user-data --disable-background-networking --disable-breakpad --disable-renderer-backgrounding --disable-hang-monitor --disable-background-timer-throttling --mute-audio --disable-web-sockets --window-size=1400,900 --no-default-browser-check --disable-first-run-ui --no-first-run --homepage=about:blank --disable-features=HttpsUpgrades --disable-direct-npapi-requests --disable-web-security --disable-notifications --disable-extensions --disable-save-password-bubble --disable-sync --headless=new about:blank'
```

For comparison, the `logging` logs before:

```
2025-02-20 14:02:57,283 9107 INFO MainThread brozzler.model.Site._accept_ssurt_if_not_redundant(model.py:282) adding ssurt com,example,//https:/ to scope accept rules
2025-02-20 14:02:57,283 9107 WARNING MainThread root.__init__(worker.py:129) not starting prometheus scrape endpoint: metrics_port is undefined
2025-02-20 14:02:57,307 9107 INFO MainThread brozzler.chrome.Chrome.start(chrome.py:230) running: '"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" -v --remote-debugging-port=9222 --remote-allow-origins=http://localhost:9222 --use-mock-keychain --user-data-dir=/var/folders/v9/fhm3y0l12zvbmtpmf39nxm6h0000gn/T/tmp8pfkm2x7/chrome-user-data --disable-background-networking --disable-breakpad --disable-renderer-backgrounding --disable-hang-monitor --disable-background-timer-throttling --mute-audio --disable-web-sockets --window-size=1400,900 --no-default-browser-check --disable-first-run-ui --no-first-run --homepage=about:blank --disable-features=HttpsUpgrades --disable-direct-npapi-requests --disable-web-security --disable-notifications --disable-extensions --disable-save-password-bubble --disable-sync --headless=new about:blank'
```

I've left the gunicorn logs from brozzler dashboard alone; those seem more difficult to hook into structlog, but also lower priority. As a result, we still configure `logging` in the same place we did before just to ensure that those remaining calls

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -184,7 +184,9 @@ class ThreadExceptionGate:
     def __enter__(self):
         assert self.thread == threading.current_thread()
         if self.pending_exception:
-            self.logger.info("raising pending exception", pending_exception=self.pending_exception)
+            self.logger.info(
+                "raising pending exception", pending_exception=self.pending_exception
+            )
             tmp = self.pending_exception
             self.pending_exception = None
             raise tmp
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -123,7 +123,7 @@ def behavior_script(url, template_parameters=None, behaviors_dir=None):
     """
     import re, json
 
-    logger = structlog.get_logger()
+    logger = structlog.get_logger(logger_name=__name__)
 
     for behavior in behaviors(behaviors_dir=behaviors_dir):
         if re.match(behavior["url_regex"], url):
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -18,6 +18,7 @@ limitations under the License.
 """
 
 import logging
+import structlog
 from pkg_resources import get_distribution as _get_distribution
 
 __version__ = _get_distribution("brozzler").version
@@ -146,7 +147,9 @@ def behavior_script(url, template_parameters=None, behaviors_dir=None):
     """
     Returns the javascript behavior string populated with template_parameters.
     """
-    import re, logging, json
+    import re, json
+
+    logger = structlog.get_logger()
 
     for behavior in behaviors(behaviors_dir=behaviors_dir):
         if re.match(behavior["url_regex"], url):
@@ -159,18 +162,18 @@ def behavior_script(url, template_parameters=None, behaviors_dir=None):
                 behavior["behavior_js_template"]
             )
             script = template.render(parameters)
-            logging.info(
-                "using template=%r populated with parameters=%r for %r",
-                behavior["behavior_js_template"],
-                json.dumps(parameters),
-                url,
+            logger.info(
+                "rendering template",
+                template=behavior["behavior_js_template"],
+                parameters=json.dumps(parameters),
+                url=url,
             )
             return script
     return None
 
 
 class ThreadExceptionGate:
-    logger = logging.getLogger(__module__ + "." + __qualname__)
+    logger = structlog.get_logger(__module__ + "." + __qualname__)
 
     def __init__(self, thread):
         self.thread = thread
@@ -181,7 +184,7 @@ class ThreadExceptionGate:
     def __enter__(self):
         assert self.thread == threading.current_thread()
         if self.pending_exception:
-            self.logger.info("raising pending exception %s", self.pending_exception)
+            self.logger.info("raising pending exception", pending_exception=self.pending_exception)
             tmp = self.pending_exception
             self.pending_exception = None
             raise tmp
@@ -198,10 +201,10 @@ class ThreadExceptionGate:
         with self.lock:
             if self.pending_exception:
                 self.logger.warning(
-                    "%r already pending for thread %r, discarding %r",
-                    self.pending_exception,
-                    self.thread,
-                    e,
+                    "exception already pending for thread, discarding",
+                    pending_exception=self.pending_exception,
+                    thread=self.thread,
+                    exception=e,
                 )
             else:
                 self.pending_exception = e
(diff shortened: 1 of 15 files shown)


brozzler/__init__.py
```diff
@@ -159,18 +136,18 @@ def behavior_script(url, template_parameters=None, behaviors_dir=None):
...
             script = template.render(parameters)
-            logging.info(
-                "using template=%r populated with parameters=%r for %r",
-                behavior["behavior_js_template"],
-                json.dumps(parameters),
-                url,
+            logger.info(
+                "rendering template",
+                template=behavior["behavior_js_template"],
+                parameters=json.dumps(parameters),
```
wilsoniya, line 142: do we need to `dumps` the parameters? Assuming later use of a json log format, deferring serialization to the logger will probably make it easier to query into `parameters`

brozzler/chrome.py
```diff
@@ -259,24 +254,21 @@ def _websocket_url(self, timeout_sec=60):
...
                 if time.time() - self._last_warning > 30:
-                    self.logger.warning(
-                        "problem with %s (will keep trying until timeout "
-                        "of %d seconds): %s",
-                        json_url,
-                        timeout_sec,
-                        e,
+                    url_logger.warning(
+                        "problem accessing url (will keep trying until timeout)",
+                        timeout=timeout_sec,
```
wilsoniya, line 270: Consider including unit:
```suggestion
                        timeout_sec=timeout_sec,
```

brozzler/cli.py
```diff
@@ -109,7 +112,50 @@ def rethinker(args):
...
+
+
 def configure_logging(args):
+    structlog.configure(
+        processors=[
+            structlog.contextvars.merge_contextvars,
+            structlog.processors.add_log_level,
+            structlog.processors.StackInfoRenderer(),
+            structlog.dev.set_exc_info,
+            structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S", utc=False),
```
wilsoniya, line 140: Would we not want timestamps in UTC?

brozzler/cli.py
```diff
@@ -371,11 +417,11 @@ def on_screenshot(screenshot_jpeg):
             on_screenshot=on_screenshot,
             enable_youtube_dl=not args.skip_youtube_dl,
         )
-        logging.info("outlinks: \n\t%s", "\n\t".join(sorted(outlinks)))
+        logger.info("outlinks: \n\t%s", "\n\t".join(sorted(outlinks)))
```
wilsoniya, line 420: Consider removing newlines and probably all escaped whitespace chars from log statements. You shouldn't need them with structured output and newlines in statements will probably make log aggregation harder and/or less effective

brozzler/__init__.py
```diff
@@ -146,7 +121,9 @@ def behavior_script(url, template_parameters=None, behaviors_dir=None):
     """
     Returns the javascript behavior string populated with template_parameters.
     """
-    import re, logging, json
+    import re, json
+
+    logger = structlog.get_logger()
```
wilsoniya, line 126: I think we probably want to pass `__name__` to `get_logger` across the board.

wilsoniya (approved):
Thanks for your work on this! Some non-blocking considerations; LGTM!

brozzler/__init__.py
mistydemeo: Good catch - we definitely don't need to do that anymore!

brozzler/__init__.py
mistydemeo: Yeah, good call. I'll update the `get_logger()` calls that are missing `logger_name` to make sure they get something appropriate.

brozzler/cli.py
mistydemeo: We probably do, agreed. This is the default structlog config - I stuck with the default formatting and timestamps until we had the chance to decide what we wanted the logs to look like.

brozzler/cli.py
mistydemeo: Good call; did that here and in the two tests that did the same thing.

```
2025-02-22 00:14:00 [info     ] outlinks                       [brozzler.cli.brozzle_page(cli.py:420)] outlinks=['https://www.iana.org/domains/example']
```

brozzler/cli.py
```diff
@@ -109,7 +112,50 @@ def rethinker(args):
     return doublethink.Rethinker(servers.split(","), db)
 
 
+# Decorates the logger name with call location, if provided
```
gretchenleighmiller, line 115: Nit: You could move this down below the function to make it a proper docstring.

```
def decorate_logger_name(a, b, event_dict):
    """Decorates the logger name with call location, if provided"""

    old_name = event_dict.get("logger_name")
```

See https://peps.python.org/pep-0257/.

gretchenleighmiller (approved):
One small non-blocking nit. Otherwise LGTM!

PR #326, opened 2025-02-25
feat: more logging options

mistydemeo:
This is a followup to #325. That PR was already long enough, and since this also adds some new CLI options, I felt it was useful breaking it out into a separate PR.

The first feature adds optional syslog prefixing. This is useful when stdout/stderr are going to the journal, and [`SyslogLevelPrefix=`](https://manpages.debian.org/testing/systemd/systemd.exec.5.en.html) is enabled. When this is on, log lines are prefixed with a three-character prefix indicating the log level (for example, `<6>` for `INFO`). systemd then strips these prefixes when ingesting them into the journal and assigns them the appropriate logging level, making it possible to do log level filtering while reading/streaming in `journalctl`.

The second feature adds an optional worker ID parameter. If specified, this value gets added to every log entry as `worker_id=$ID`, making it easier to filter logs by the worker that created them.


brozzler/cli.py
```diff
@@ -131,30 +142,58 @@ def decorate_logger_name(a, b, event_dict):
     return event_dict
 
 
+# https://manpages.debian.org/testing/libsystemd-dev/sd-daemon.3.en.html
+def systemd_log_prefix(_, level, log):
```
wilsoniya, line 146: consider indicating private visibility: 

```suggestion
def _systemd_log_prefix(_, level, log):
```

Admittedly nittish considering it's very unlikely anyone's going to reuse this function outside the context of logging, but I always like to make it perfectly clear that one would be doing something wrong using a private symbol outside of the enclosing module scope.

mistydemeo: Good call!

brozzler/cli.py
```diff
@@ -131,30 +142,58 @@ def decorate_logger_name(a, b, event_dict):
...
+                structlog.processors.CallsiteParameter.FUNC_NAME,
+                structlog.processors.CallsiteParameter.LINENO,
+            ],
+        ),
+        decorate_logger_name,
+        structlog.dev.ConsoleRenderer(),
+    ]
+
+    if args.syslogd_log_prefix:
+        processors.append(_systemd_log_prefix)
```
gretchenleighmiller, line 183: Asking out of ignorance more than anything: This is _supposed_ to be the function itself passed in, not an invocation of the function, right? e.g. it shouldn't be `processors.append(_systemd_log_prefix())`

gretchenleighmiller (approved):
One sanity-check question that's more for my understanding. Otherwise LGTM 👍

brozzler/cli.py
mistydemeo: That's right! The processor chain is a chain of callables that are applied to the log data in sequence in order to produce the output.

brozzler/cli.py
```diff
@@ -74,6 +74,17 @@ def add_common_options(arg_parser, argv=None):
         const=logging.DEBUG,
         help=("very verbose logging"),
     )
+    arg_parser.add_argument(
+        "--syslogd-log-prefix",
+        dest="syslogd_log_prefix",
+        action="store_true",
+        help="add syslogd log level prefix for journalctl filtering",
+    )
+    arg_parser.add_argument(
```
galgeek, line 83: already noted in the ait-ansible MR — if we need a separate worker-id CLI var, I'd prefer to default the metrics-port var to that, instead of requiring what are now identical values to be set twice

mistydemeo: I'm wondering if these two values are always the same thing for other users? It feels like we might be leaning a little too hard into optimizing this around our own deployment in a way that wouldn't work for other users.

galgeek: metrics-port was only introduced in the past couple of months — I'd be surprised if it affected any other users

mistydemeo: Yes, true! That makes sense. That said, I feel like it's probably easier to separate those values now rather than at some point in the future after they've been used for awhile.

PR #328, opened 2025-02-25
chore: use ruff for formatting

mistydemeo:
Had a discussion about this with Alex; we thought it might be a good idea to try out `ruff` for formatting rather than `black`. Seems to be working well so far in local testing.

There are a few minor changes here compared to `black`; it flagged unnecessary string concatenations that `black` didn't care about, and has slightly different opinions on line length.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: format
 format:
-	venv/bin/black -t py35 -t py36 -t py37 -t py38 -t py39 -t py310 -t py311 -t py312 .
+	venv/bin/ruff format --target-version py37 .
 
 .PHONY: ck-format
 ck-format:
-	venv/bin/black --check .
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: format
 format:
-	venv/bin/black -t py35 -t py36 -t py37 -t py38 -t py39 -t py310 -t py311 -t py312 .
+	venv/bin/ruff format --target-version py37 .
```
galgeek, line 3: Python 3.7 is the earliest version we can target, per
https://docs.astral.sh/ruff/settings/#target-version

mistydemeo: I should have mentioned, by the way, that unlike `black` it doesn't want you to specify every version you support - just the lowest.

PR #329, opened 2025-02-26
test: add CI

mistydemeo:
This adds two CI runs: a quick one that happens for every pull request and merge to master, and a longer one that happens daily. The long one, in my tests so far, took about six hours(!) - so it's not feasible for us to do on every pull request.

The pull request/master merge tests are all reliably passing right now, but the full suite isn't. I'm not clear if these are legitimate failures, or if it's a sign that some of the code it's testing has changed since it was written. Here's the full test run I ran on it overnight: https://github.com/mistydemeo/brozzler/actions/runs/13534089995/job/37822403796

I added a new installation group to setup.py because the `easy` group isn't currently installable, and some of the dependencies specified there need to be present for the tests to run.

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -1,0 +1,22 @@
+name: Full test suite
+
+on:
+  schedule:
+    - cron: "0 6 * * *" # 10PM Pacific daily
+
+jobs:
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - uses: ./.github/workflows/setup
+
+      - name: Run tests
+        run: |
+          py.test --tb=native --verbose tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,28 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    branches:
+      - main
+      - master
+
+jobs:
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - uses: ./.github/workflows/setup
+
+      - name: Run tests
+        run: |
+          py.test --tb=native --verbose tests/test_cli.py tests/test_units.py
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,12 +14,15 @@ jobs:
   test:
     name: Run tests
     runs-on: ubuntu-latest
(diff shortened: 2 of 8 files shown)


.github/workflows/daily.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.8'
```
galgeek, line 16: Could we run under Python 3.12?
AIT brozzler is running under Python 3.12 now

.github/workflows/tests.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.8'
```
galgeek, line 22: again, let's run under Python 3.12 if we can

wilsoniya:
I'm not familiar w/ gh actions, so I don't have much valuable feedback with respect CI configs. 

Consider creating a ticket for fixing/updating/deleting the broken tests and linking to said ticket near the `xfail` decoration.

.github/workflows/daily.yaml
gretchenleighmiller: We can do a matrix of versions, right? So I think at a minimum 3.8 and 3.12 would be good targets, but I would prefer we don't _just_ test 3.12. There are likely Brozzler users running under 3.8.

mistydemeo: Yes, true! It shouldn't significantly affect our buildtime to test on both. If we do that, I'm assuming 3.8 is an appropriate floor? Our pyproject.toml claims that's the lowest supported version.

gretchenleighmiller: 3.8 is the most recent version of Python that's still supported (although the last security update was October 2024), so I think it's very reasonable to say that we're going to make the cutoff the oldest supported version of Python.

mistydemeo: All right, after fixing a couple bugs introduced by the swap to 3.12, I've added a matrix that runs under both 3.8 and 3.12.

brozzler/__init__.py
```diff
@@ -19,9 +19,9 @@
 
 import logging
 import structlog
-from pkg_resources import get_distribution as _get_distribution
+from importlib.metadata import version as _version
 
-__version__ = _get_distribution("brozzler").version
+__version__ = _version("brozzler")
```
mistydemeo, line 24: I should note that, like the f-strings that we've added recently, this sets a floor on our Python version - importlib was introduced in 3.8. Our pyproject.toml already declared 3.8 as our minimum Python version though.

mistydemeo:
Now that I've made the changes to add the 3.8/3.12 matrix, happy with me merging this?

PR #332, opened 2025-03-03
Disable OptimizationHints,OptimizationGuideModelDownloading

vbanos:
Chrome downloads 80MB+ optimisation_guide_model_store by default. We don't need that.


vbanos:
optimisation-guide-chrome

I was looking into Chrome temp files and I noticed this.
The options I'm suggesting to use prevent downloading it.

vbanos:
Duplicate of https://github.com/internetarchive/brozzler/pull/320

PR #334, opened 2025-03-04
__init__.py: rework imports

mistydemeo:
Although doublethink is an optional dependency to allow brozzler to be used as a library without it, in practice we had some mandatory import statements that prevented brozzler from being imported without it. This fixes that by gating off some of the imports and exports.

If doublethink is available, brozzler works as it is now. But if it isn't, we make a few changes:

* `brozzler.worker`, `brozzler.cli` and `brozzler.model` reexports are disabled
* One `brozzler.cli` function, which is used outside brozzler's own cli, has been moved into brozzler's `__init__.py`. For compatibility, it's reexported from `brozzler.cli`.

PR #335, opened 2025-03-04
Make doublethink and yt-dlp truly optional

avdempsey:
While working on updating Umbra, Misty uncovered Brozzler main `__init__.py` unconditionally imports doublethink and yt-dlp, even though these dependencies are listed as optional.

The requirement to import doublethink is resolved by using the standard library to create a timezone aware date time object for the moment of the Unix epoch (that specific second in 1970). It also moves off of the deprecated use of `datetime.utcfromtimestamp`.

The requirement to import yt-dlp is resolved by moving the `should_ytdlp` method to the worker class itself, and conditionally import the `ydl` module if we are going to use it.

I considered moving things from the main `brozzler.__init__.py` but I was concerned that could break code that's not visible to me.

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -321,10 +321,9 @@ def _remove_query(url):
 # XXX chop off path after last slash??
 site_surt_canon = urlcanon.Canonicalizer(urlcanon.semantic.steps + [_remove_query])
 
-import doublethink
 import datetime
 
-EPOCH_UTC = datetime.datetime.utcfromtimestamp(0.0).replace(tzinfo=doublethink.UTC)
+EPOCH_UTC = datetime.datetime.fromtimestamp(0.0, tz=datetime.timezone.utc)
 
 # we could make this configurable if there's a good reason
 MAX_PAGE_FAILURES = 3
--- a/brozzler/worker.py
+++ b/brozzler/worker.py
@@ -39,7 +39,6 @@ import urlcanon
 from requests.structures import CaseInsensitiveDict
 import rethinkdb as rdb
 from . import metrics
-from . import ydl
 
 r = rdb.RethinkDB()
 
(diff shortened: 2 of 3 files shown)


brozzler/__init__.py
```diff
@@ -321,10 +321,9 @@ def _remove_query(url):
 # XXX chop off path after last slash??
 site_surt_canon = urlcanon.Canonicalizer(urlcanon.semantic.steps + [_remove_query])
 
-import doublethink
```
mistydemeo, line 324: We still end up transiently pulling in doublethink from importing `brozzler.worker`, `brozzler.model` and functions from `brozzler.cli`. We need to do something like #334 to gate off importing anything from those models if `import doublethink` is unavailable. eg, with this branch and doublethink not available:

```
>>> import brozzler
  File "<stdin>", line 1, in <module>
  File "/Users/mistydemeo/git/brozzler/brozzler/__init__.py", line 331, in <module>
    from brozzler.worker import BrozzlerWorker
  File "/Users/mistydemeo/git/brozzler/brozzler/worker.py", line 36, in <module>
    import doublethink
ModuleNotFoundError: No module named 'doublethink'
```

mistydemeo (changes requested):
Good call on the ydl stuff! That cleans that up nicely. We need a little more for doublethink though.

brozzler/worker.py
```diff
@@ -260,6 +259,38 @@ def thumb_jpeg(self, full_jpeg):
         img.save(out, "jpeg", quality=95)
         return out.getbuffer()
 
+    def should_ytdlp(self, logger, site, page, page_status, skip_av_seeds):
```
mistydemeo, line 262: Just had a thought, is it also worth adding an attempt to import ydl and erroring out if it can't? eg, we shouldn't ytdlp if the optional extra is missing.

```python
try:
    import yt_dlp
except ImportError:
    return False
```

avdempsey: Oh good idea, probably with log message

mistydemeo: Done in a0a06b5d3316557830dcbc480da47d5ac61ad17a in the other PR.

mistydemeo:
The content of this is in #334 now, so I think we can close this.

PR #340, opened 2025-03-06
Makefile improvements + pre-commit hook

gretchenleighmiller:


.github/workflows/python-formatting.yml
```diff
@@ -20,12 +20,12 @@ jobs:
       with:
         python-version: '3.8'
     - name: Create virtual environment
-      run: python -m venv venv
+      run: python -m venv .venv
```
mistydemeo, line 23: I'm very in favour of moving the venv to `.venv`, but let's add that path to the gitignore. (Leave `venv` for now so existing virtualenvs stay ignored.)

gretchenleighmiller: Good call! Done.

Makefile
```diff
@@ -1,7 +1,50 @@
...
+.git/hooks/pre-commit:
+	ln -s $(realpath ./dev/pre-commit) $@
+
+.PHONY: check
+check:
+	$(VIRTUAL_ENV_DIR)/bin/ruff check --target-version py37 .
+
+.PHONY: check-format
+check-format:
+	$(VIRTUAL_ENV_DIR)/bin/ruff check --select I
```
mistydemeo, line 44: Is this an abbreviated name, or is `I` the entire official name for this check?

gretchenleighmiller: I believe `I` is the actual rule name rather than an abbreviation. It's selecting all `isort` rules.

Something we could do is specify this in `pyproject.toml`. See https://docs.astral.sh/ruff/linter/#rule-selection

mistydemeo (commented):
Looking good to me, but right now `make check` is failing for me - if we're introducing a new check and adding it to a precommit hook, let's make sure it passes as-is first.

gretchenleighmiller:
> Looking good to me, but right now `make check` is failing for me - if we're introducing a new check and adding it to a precommit hook, let's make sure it passes as-is first.

Fair! But...
```
Found 554 errors.
[*] 96 fixable with the `--fix` option (185 hidden fixes can be enabled with the `--unsafe-fixes` option).
make: *** [check] Error 1
```

Let's do a separate branch addressing those issues. I think it's fine to `noqa` the tricky bits for now; at least that would document them. What do you think?

mistydemeo:
Yeah, sounds good! No need to fix them imo, just figure if we're adding a hook encouraging people to `make check`, that hook should work.

mistydemeo (approved):
`make check` should be fine now that #343 is merged!

PR #343, opened 2025-03-07
ruff linting fixes

gretchenleighmiller:
Linting fixes based on running this exact invocation: `ruff check --target-version py37 .`. Also ran `ruff format --target-version py37 .` and `ruff check --target-version py37 --select I . --fix` to ensure proper formatting and import sorting was intact.

The fixes were manual, but largely mechanical and tried to be as unobtrusive as possible. There are many more opportunities for improvement that can come in subsequent passes.

### DIFF
--- a/brozzler/__init__.py
+++ b/brozzler/__init__.py
@@ -25,9 +25,6 @@ from importlib.metadata import version as _version
 import structlog
 import urlcanon
 
-from brozzler.browser import Browser, BrowserPool, BrowsingException
-from brozzler.robots import is_permitted_by_robots
-
 __version__ = _version("brozzler")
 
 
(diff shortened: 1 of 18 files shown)


brozzler/__init__.py
```diff
@@ -403,14 +401,8 @@ def suggest_default_chrome_exe():
     return "chromium-browser"
 
 
-import datetime
-
 EPOCH_UTC = datetime.datetime.fromtimestamp(0.0, tz=datetime.timezone.utc)
 
-
-from brozzler.browser import Browser, BrowserPool, BrowsingException
-from brozzler.robots import is_permitted_by_robots
```
mistydemeo, line 412: This is imported here to rexport it via the `__all__` list below. Ditto the `brozzler.browser` imports.

mistydemeo (commented):
Looks good! Just one comment.

brozzler/__init__.py
gretchenleighmiller: Fixed! Thank you!

mistydemeo (approved):

PR #347, opened 2025-03-22
chore: annotate claim_sites()

adam-miller:


gretchenleighmiller (approved):

gretchenleighmiller:
I'm comfortable with adding these comments to clarify behavior. One last thing is that you need to do a `ruff` pass over the file so it passes CI; I'm guessing it doesn't like how the comments are formatted.

PR #348, opened 2025-03-27
feat: Create new claim_sites() query, and fix frontier tests

adam-miller:


brozzler/frontier.py
```diff
@@ -16,6 +16,9 @@
 limitations under the License.
 """
 
+import datetime
+from typing import List, Dict
```
gretchenleighmiller, line 20: These are deprecated as of Python 3.9; you can use just use `list` and `dict`. Once we drop Python 3.8 support (and it might be good to do that soon considering it's reached EOL), we can safely remove these imports.

brozzler/__init__.py
```diff
@@ -398,7 +399,7 @@ def suggest_default_chrome_exe():
     return "chromium-browser"
 
 
-EPOCH_UTC = datetime.datetime.fromtimestamp(0.0, tz=datetime.timezone.utc)
+EPOCH_UTC = datetime.datetime.fromtimestamp(0.0, tz=doublethink.UTC)
```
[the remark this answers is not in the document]
adam-miller: This would break all of our tests because rethink returns its own tz object. Maybe I can find another way around that.

brozzler/frontier.py
```diff
@@ -30,6 +33,57 @@ class UnexpectedDbResult(Exception):
...
+) -> List[str]:
+    job_counts = {}
+    claimable_sites = []
+    now = datetime.datetime.now(datetime.timezone.utc)
+
+    for site in active_sites:
+        is_claimable = False
+
+        # If site not claimed and not disclaimed within last 20 seconds
+        if not site["claimed"] and site.get("last_disclaimed", 0) <= (
```
mistydemeo, line 47: Is `"claimed"` guaranteed to be in `site`? I noticed you're guarding against all of the other keys, but not this one.

mistydemeo (commented):
I gave this a test, and looks like the frontier tests pass for me! Maybe we can add those to the list in `.github/workflow/tests.yml` so they get run regularly?

brozzler/frontier.py
```diff
@@ -101,68 +155,35 @@ def _vet_result(self, result, **kwargs):
                         "expected %r to be %r in %r" % (k, expected, result)
                     )
 
-    def claim_sites(self, n=1):
+    def get_active_sites(self) -> List[Dict]:
+        active_sites = (
```
galgeek, line 159: is it vital to iteratie over all active sites every time?

or is there much efficiency to be gained by re-checking already claimed sites only when there are no unclaimed sites to be claimed? I guess this assumes that rethinkdb can create a claimed sites by job dict more efficently than we can in Python

avdempsey: There's a level of scaling where we definitely won't want to pull down all active sites. I think for now it will be easier for us to get the logic right by pulling everything down, and we have more Brozzler Workers than we do RethinkDBs, so I'm ok with a little extra work in Python land.

brozzler/frontier.py
```diff
@@ -30,6 +33,57 @@ class UnexpectedDbResult(Exception):
...
+        is_claimable = False
+
+        # If site not claimed and not disclaimed within last 20 seconds
+        if not site["claimed"] and site.get("last_disclaimed", 0) <= (
+            now - datetime.timedelta(seconds=20)
+        ):
+            is_claimable = True
+
+        # or site has been disclaimed more than an hour ago
+        if "last_claimed" in site and site["last_claimed"] <= (
```
galgeek, line 53: could this be an `else` after the previous `if`?
I don't think we want to do this check when the first `if` matches.

brozzler/frontier.py
```diff
@@ -30,6 +33,57 @@ class UnexpectedDbResult(Exception):
...
+        ):
+            is_claimable = True
+
+        # Count number of claimed sites per job_id (optional field)
+        if site["claimed"] and "max_claimed_sites" in site and "job_id" in site:
+            job_id = site["job_id"]
+            job_counts[job_id] = job_counts.get(job_id, 0) + 1
+
+        if is_claimable:
+            claimable_sites.append(site)
```
galgeek, line 64: could we stop looping here, if we we've got "enough" claimable sites,
and if we've counted all "claimed" sites?

is "enough" maybe 10?

brozzler/frontier.py
```diff
@@ -30,6 +33,57 @@ class UnexpectedDbResult(Exception):
...
+
+    site_ids_to_claim = []
+    # gather sites that are under the max without going over
+    for site in claimable_sites:
+        if (
+            "max_claimed_sites" in site
+            and "job_id" in site
+            and job_counts.get(site["job_id"], 0) < site["max_claimed_sites"]
+        ):
+            site_ids_to_claim.append(site["id"])
```
galgeek, line 74: we could do this before the `if` (and include only the job_counts update after `if`, and skip the second `if` block)

brozzler/chrome.py
```diff
@@ -222,6 +222,7 @@ def start(
             chrome_args.append("--ignore-certificate-errors")
         if proxy:
             chrome_args.append("--proxy-server=%s" % proxy)
+        self.logger.info(f"Chrome launched with args {chrome_args} proxy is {proxy}")
```
mistydemeo, line 225: Is this intended to be a temporary debug statement before this is merged, or permanent? If the latter, you can probably just combine this with the logging statement two lines down by adding `proxy=proxy` to the logger dictionary.

brozzler/frontier.py
gretchenleighmiller: Logically it's fine as two separate clauses. The result is the same if either of the conditions are true. They could be combined into one big `if` with an `or`, i.e. it would short-circuit if the first condition evaluates to true and save a few cycles, but I think it's probably clearer to separate them out and annotate them with comments.

brozzler/frontier.py
adam-miller: Yes, good to double check that. There is a default of False, and also the input is a result of a query filtering on that field, so it has to exist at this point.

brozzler/frontier.py
adam-miller: Correct. It's a big or. The complexity with the large number of conditionals made me feel it's easier to read as separate clauses.

brozzler/chrome.py
adam-miller: It was useful in debugging a failing test. Probably worth keeping, so I'll add this to the other log line.

brozzler/frontier.py
adam-miller: These 2 if clauses are also effectively a large `or`. Merging them in this way fails some test cases where we exceed max claimed sites when it is provided

brozzler/frontier.py
adam-miller: Indeed. We can break early if `len(claimable_sites) >= max_sites_to_claim`

brozzler/frontier.py
```diff
@@ -30,6 +33,62 @@ class UnexpectedDbResult(Exception):
...
 
 
+def filter_claimable_site_ids(
+    active_sites: List[Dict],
+    reclaim_cooldown: int,
+    max_sites_to_claim=1,
+) -> List[str]:
+    job_counts = {}
+    claimable_sites = []
+    now = datetime.datetime.now(datetime.timezone.utc)
```
avdempsey, line 43: This is technically correct, but for fun I like to reduce the repetition of "datetime" and spell it like this:

```
from datetime import datetime, UTC

now = datetime.now(UTC)
```

gretchenleighmiller: If you're gonna do this, you need to add `timedelta` as well, because it's used in this file. I usually prefer just doing `import datetime` because `timedelta` and several other common utilities and classes are under the top-level `datetime`, and it's clearest to me to start from the top of the package given the `datetime` and `datetime.datetime` confusion.

avdempsey: Oops, datetime.UTC is added in py3.11, so never mind

PR #351, opened 2025-05-08
dependabot: add yt-dlp in qa

mistydemeo:
Also sets up an auto-approve for dependabot PRs.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,21 @@
+name: Dependabot
+on: pull_request
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependabot:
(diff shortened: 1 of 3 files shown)


galgeek (commented):
looks good to me, for yt-dlp — I do wonder how we'd blacklist a particular version, if desired

mistydemeo:
I think Dependabot should respect the version specifier in `setup.py`, so if we mark the version as "greater than X version, lower than Y" it shouldn't upgrade us to Y or newer.

mistydemeo:
I've added the script that automatically performs a test with five videos; if at least three pass, we go ahead with the auto-merge to the `qa` branch.

galgeek (commented):
looks good to me

PR #363, opened 2025-05-28
ci: try installing chrome before uv sync

mistydemeo:
Looks like we get an incompatible version of Chrome unless we explicitly install it running the test.


mistydemeo:
Added a commit to merge after approving. I meant to do this originally and just forgot.

PR #366, opened 2025-06-04
deps: specify more extras for yt-dlp

mistydemeo:
I closed #360, but still a good idea to get the additional extras in.


galgeek (approved):
looks good to me

PR #367, opened 2025-06-09
increase HEADER_REQUEST_TIMEOUT

galgeek:


vbanos:
I think that 30 is already too much.
In production we always use values < 10 from my experience.
Why increase it to 60?

galgeek:
Thanks, @vbanos!  It's perhaps more important for our application to get the page header, at the expense of a larger timeout.  I'll keep your recommendation in mind, for a possible revision.

PR #373, opened 2025-06-16
Add `headless` option to Chrome.start

TheTechRobo:
Previously, Chrome's headless mode was always turned on. This makes it optional when using Brozzler as a module. (If there's interest in a command-line option for this, I can add that too.)

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -146,6 +146,7 @@ class Chrome:
         websocket_timeout=60,
         window_height=900,
         window_width=1400,
+        headless=True,
     ):
         """
         Starts chrome/chromium process.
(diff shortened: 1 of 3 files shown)


mistydemeo:
I'm curious, what's the benefit of having this as an option? Is there something this lets it capture that wouldn't be possible otherwise? I'd be interested in getting it documented somewhere if we add it.

TheTechRobo:
@mistydemeo It's useful for debugging. I also wouldn't be surprised if headless Chrome is easier to fingerprint, but I haven't looked into that.

@galgeek This project in particular requires hooking into brozzler to allow for some custom logic, which a CLI can't really emulate.

galgeek:
It would be good to be able to disable headless via brozzler's CLI.

There's an example of another CLI addition here:
https://github.com/internetarchive/brozzler/pull/307/files

mistydemeo:
Just wanted to check back in, are you up for adding a CLI flag for this? If not, I'm happy to merge and add that myself in a followup.

TheTechRobo:
Sorry, I meant to get back to this earlier. I've pushed a commit to add an option to the CLI. Sorry about that!

mistydemeo:
No worries at all! I appreciate it.

PR #374, opened 2025-06-21
predup youtube playlists and tabs, and more...

galgeek:


pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
   "python-magic>=0.4.15",
   "prometheus-client>=0.20.0",
   "structlog>=25.1.0",
+  "psycopg[binary]>=3.2.6",
```
avdempsey, line 38: This should be an optional dependency.

brozzler/ydl.py
```diff
@@ -418,6 +420,43 @@ def _try_youtube_dl(worker, ydl, site, page):
...
+                "SELECT containing_page_url from video where seed = %s and containing_page_url like %s",
+                (seed, containing_page_url_pattern),
+            )
+        else:
+            return None
+        with psycopg.connect(VIDEO_DATA_SOURCE) as conn:
+            with conn.cursor(row_factory=psycopg.rows.scalar_row) as cur:
+                cur.execute(pg_query)
+                return cur.fetchall()
+    return None
```
avdempsey, line 457: We should consider wrapping this call with an interface as a client
- This would allow us to have the client use pg directly or make an API call
- A client could maintain connections instead re-connecting to PG on every check
- We need to add unit tests before we ship this feature, a client might make unit testing easier

brozzler/ydl.py
```diff
@@ -418,6 +420,43 @@ def _try_youtube_dl(worker, ydl, site, page):
...
+    if not VIDEO_DATA_SOURCE:
+        return None
+
+    if VIDEO_DATA_SOURCE and VIDEO_DATA_SOURCE.startswith("postgresql"):
+        import psycopg
+
+        account_id = site.account_id if site.account_id else None
+        seed = site.metadata.ait_seed_id if site.metadata.ait_seed_id else None
+        if source == "youtube":
+            containing_page_url_pattern = "http://youtube.com/watch"  # yes, video data canonicalization uses "http"
```
avdempsey, line 433: Does psycopg implicitly add "%" to the end of a LIKE pattern? I would expect something like "http://youtube.com/watch/%". This should be covered by a test.

brozzler/ydl.py
```diff
@@ -418,6 +420,43 @@ def _try_youtube_dl(worker, ydl, site, page):
     return ie_result
 
 
+def get_video_captures(site, source="youtube"):
```
avdempsey, line 423: Can we type hint this function so we know what it returns? Maybe it should return an empty list instead of None when we don't find anything.

brozzler/ydl.py
```diff
@@ -37,15 +41,111 @@
...
+
+        if account_id and seed and source:
+            pg_query = (
+                "SELECT distinct(containing_page_url) from video where account_id = %s and seed = %s and containing_page_url like %s",
+                (
+                    account_id,
+                    seed,
+                    containing_page_url_pattern,
+                ),
+            )
```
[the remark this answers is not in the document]
galgeek: this lookup is to enable checks on multiple youtube channel and playlist URLs before they're added to outlinks

brozzler/ydl.py
```diff
@@ -444,10 +544,20 @@ def do_youtube_dl(worker, site, page, ytdlp_proxy_endpoints):
...
+                )
+                uncaptured_youtube_watch_pages = []
+                for e in ie_result.get("entries_no_dl", []):
+                    youtube_watch_url = f"https://www.youtube.com/watch?v={e['id']}"
+                    if youtube_watch_url in captured_youtube_watch_pages:
+                        continue
+                    uncaptured_youtube_watch_pages.append(youtube_watch_url)
+                if uncaptured_youtube_watch_pages:
+                    outlinks.add(uncaptured_youtube_watch_pages)
+        # todo: handle outlinks for instagram and soundcloud, other media source, here (if anywhere)
```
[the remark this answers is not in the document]
galgeek: thanks, Alex!  updated

brozzler/ydl.py
galgeek: recent updates include most of this work

PR #377, opened 2025-06-25
tests: use github-format in ci

mistydemeo:


Makefile
```diff
@@ -37,14 +48,14 @@ clean: $(BROZZLER_EGG_LINK)
 
 .PHONY: check
 check:
-	$(VIRTUAL_ENV_DIR)/bin/ruff check --target-version py37 .
+	$(VIRTUAL_ENV_DIR)/bin/ruff check $(OUTPUT_FLAGS) --target-version py37 .
```
gretchenleighmiller, line 51: It seems that we could completely remove `--target-version` from the Ruff invocations. Ruff can fall back to `requires-python` as specified in pyproject.toml.

See https://docs.astral.sh/ruff/settings/#target-version.

Currently our pyproject.toml targets 3.9 or higher.
```
requires-python = ">=3.9"
```

mistydemeo: Done.

PR #378, opened 2025-06-26
Video exclusion: skip YouTube UMP packets too

mistydemeo:
In testing a page with an embedded YouTube video with video exclusion enabled, I found that brozzler ended up capturing about 30MB of [UMP packets](https://github.com/gsuberland/UMP_Format/blob/main/UMP_Format.md). We should be filtering those out too.

### DIFF
--- a/brozzler/worker.py
+++ b/brozzler/worker.py
@@ -412,8 +412,10 @@ class BrozzlerWorker:
         Determines if the page's Content-Type header specifies that it contains
         a video.
         """
-        return (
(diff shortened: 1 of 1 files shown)


galgeek (changes requested):
I believe we'd rather not capture `application/vnd.yt-ump` content — it does not contribute to replay, and is often quite large.

This content may nevertheless be a clear indication that there's a video on the page.

mistydemeo:
Agreed! This is making sure we skip it when video exclusion is enabled. This function is only used to decide whether something should be skipped or not.

PR #385, opened 2025-07-18
remove unnecessary imports, use find_spec

mistydemeo:
This was flagged by ruff check - if we just want to find out if a package is available, and don't need to actually import it, we can use importlib.util.find_spec() to resolve it. This can lead to a moderate speedup too, since the import might be slow.


mistydemeo:
Caught one more, so pushed that one too. The `F401` checks after that ruff missing that the imports are being added to `__all__`.

PR #394, opened 2025-08-10
Track network activity and wait for idle when visiting hashtags

TheTechRobo:
In current brozzler, each anchor takes 5 hardcoded seconds to visit. On sites with a lot of anchors, especially static ones where no wait is actually necessary, this quickly blows up crawling time.

This PR starts tracking network activity in the websocket thread. `visit_hashtags` can then wait for the network to be idle on each anchor instead of the blind delay.

Non-scientific example (using `--no-headless` so you can see the improvement):

```
# 940dadfc
$ time brozzle-page https://constitution.congress.gov/browse/essay/amdt14-S1-5-6-2/ALDE_00013768/ --no-headless
brozzle-page  --no-headless  8.16s user 2.06s system 6% cpu 2:38.82 total

# this PR
$ time brozzle-page https://constitution.congress.gov/browse/essay/amdt14-S1-5-6-2/ALDE_00013768/ --no-headless
brozzle-page  --no-headless  8.01s user 2.56s system 19% cpu 54.775 total
```

I also observed speedups on websites that do load content on anchors, like https://www.army.mil/heroes/, since it usually doesn't take more than five seconds to load.

The new logic for each hashtag is to wait one second (to ensure any JavaScript event handlers have fired) and then wait for at least two seconds of idle. It will not wait for more than 4 seconds, so each hashtag will take either the same 5 seconds total that it does now or less. Let me know if you'd like those values changed.

Notes:

- This adds a lock, since the attributes could be updated by the websocket thread at any time.
- There is an ugly `_should_track_request` check added due to https://github.com/GoogleChrome/lighthouse/issues/11850 (inspired by https://github.com/GoogleChrome/lighthouse/pull/11851/files).
- A bunch of logic was added to `_handle_message`. Let me know if there are any style issues or you'd like me to factor more of it out.
- I've written a test for this. Not very experienced with pytest so I don't know how well it's written.

Happy to make any adjustments and changes as necessary. :-)


mistydemeo (approved):
Makes sense; thank you for the detailed explanation and for adding a test! I'll look to get this tested out a bit prior to merging, but I think this looks good.

mistydemeo:
Thanks for your patience - been taking some time to test, and it's all looking good. Thank you again! I really appreciate the contribution. We'll get this in the next release.

PR #426, opened 2025-12-04
Use chrome -v only when running in DEBUG logging level

vbanos:
We always use `chrome -v` but `Chrome._read_stderr_stdout()` is processing chrome verbose stdout/stderr outputs and logs them only using `logger.debug`. If not, they are ignored.

We change Chrome init to use `-v` only when the logging level is `DEBUG` to save resources.

### DIFF
--- a/brozzler/chrome.py
+++ b/brozzler/chrome.py
@@ -17,6 +17,7 @@ limitations under the License.
 """
 
 import json
+import logging
 import os
 import re
 import select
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Makes sense! Will there be any visible change to users?

vbanos:
No, there wasn't any visible change to users.
STDOUT/STDERR proces outputs are only used by the developers to see potential issues when running in `DEBUG` logging level.

PR #475, opened 2026-05-12
do not count stale site claims against max_claimed_sites cap

avdempsey:
Crashing workers or RethinkDB problems can result in orphaned site claims. The brozzler worker doesn't disclaim the site, and the claim goes stale. Jobs with more than `max_active_sites` stale site claims permanently deadlock.

The frontier's `filter_claimable_site_ids` function already had a staleness check, but then it counted stale claims towards the job's current total.

This PR changes `filter_claimable_site_ids` to not count stale site claims in a `job_counts`.

### DIFF
--- a/brozzler/frontier.py
+++ b/brozzler/frontier.py
@@ -51,14 +51,23 @@ def filter_claimable_site_ids(
         ):
             is_claimable = True
 
-        # or site has been disclaimed more than an hour ago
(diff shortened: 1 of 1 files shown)


avdempsey:
This duplicated work done by @adam-miller. Closing in favor of https://github.com/internetarchive/brozzler/pull/475

